PR #6, opened 2023-09-19
Update README.md and fix SQL repository access layer to use table prefix

braulio-dev:
README.md:
* Add development setup section for interested contributors
* Change spigot version from 1.19 -> 1.20.1. The project's spigot dependency is the latest version.
* Add MythicMobs as a dependency

SQL access layer (Specifically ClientRepository.java)
* Add use of table prefixes as per the rest of the project.

SQL migration files (Specifically in champions/src/main/resources/champions-migrations):
* Add use of table prefixes in reference to core module tables as per the rest of the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,7 @@ Repository containing all of BetterPvP's plugins for 1.20.1.
     ```bash
     git clone https://github.com/Mykindos/BetterPvP.git
     ```
-2. Compile the project using the Gradle task `reobfJar` to properly shade all project dependencies:
-   **NOTE: You must run `reobfJar` to properly obfuscate the Mojang mappings.**
+2. Compile the project using the Gradle task `assemble` to properly shade all project dependencies:
     ```bash
     ./gradlew assemble
     ```
--- a/README.md
+++ b/README.md
@@ -1,12 +1,36 @@
-Repository containing all of BetterPvP's plugins for 1.19.
+Repository containing all of BetterPvP's plugins for 1.20.1.
 
 ## Dependencies:
 - Java 17
 - PaperSpigot
 - A MySQL server
+- [MythicMobs](https://www.spigotmc.org/resources/mythicmobs.5702)
 - [ProtocolLib](https://www.spigotmc.org/resources/protocollib.1997)
 - [LibsDisguises](https://www.spigotmc.org/resources/libs-disguises-free.81)
 
 
-## Setup
-Instructions will become available later
+## Development Setup
+1. Clone this repository and open it with your IDE of choice:
+    ```bash
+    git clone https://github.com/Mykindos/BetterPvP.git
+    ```
+2. Compile the project using the Gradle task `reobfJar` to properly shade all project dependencies:
+   **NOTE: You must run `reobfJar` to properly obfuscate the Mojang mappings.**
+    ```bash
+    ./gradlew reobfJar
+    ```
+   
+3. Copy the generated jar files from all modules (`./build/libs/`) into your server's `plugins` folder.
+4. In your plugins folder, make sure to also include ProtocolLib, LibsDisguises, and MythicMobs as they are required dependencies.
+5. Create a MySQL database and import all `.sql` files located in the resource folders of each module.
+6. Start your server and wait for all plugins to load. You should get an error saying that the server couldn't establish a connection to the database.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -1,12 +1,36 @@
...
 
-## Setup
-Instructions will become available later
\ No newline at end of file
+## Development Setup
+1. Clone this repository and open it with your IDE of choice:
+    ```bash
+    git clone https://github.com/Mykindos/BetterPvP.git
+    ```
+2. Compile the project using the Gradle task `reobfJar` to properly shade all project dependencies:
```
Mykindos, line 17: You just need to run 'assemble', this will run reobj jar by default

README.md
```diff
@@ -1,12 +1,36 @@
...
+   **NOTE: You must run `reobfJar` to properly obfuscate the Mojang mappings.**
+    ```bash
+    ./gradlew reobfJar
+    ```
+   
+3. Copy the generated jar files from all modules (`./build/libs/`) into your server's `plugins` folder.
+4. In your plugins folder, make sure to also include ProtocolLib, LibsDisguises, and MythicMobs as they are required dependencies.
+5. Create a MySQL database and import all `.sql` files located in the resource folders of each module.
```
Mykindos, line 25: You only need to create the database, the rest is handled by flyway

database.getConnection().runDatabaseMigrations(getClass().getClassLoader(), "classpath:champions-migrations", databasePrefix);

^ above code is in onEnable for each plugin

This is also where the ${tablePrefix} gets populated when executing the SQL files, so your changes to the sql files will break this automatic feature

core/src/main/java/me/mykindos/betterpvp/core/client/repository/ClientRepository.java
```diff
@@ -38,7 +38,7 @@ public ClientRepository(Database database) {
     @Override
     public List<Client> getAll() {
         List<Client> clients = new ArrayList<>();
-        String query = "SELECT * FROM clients;";
+        String query = "SELECT * FROM " + databasePrefix + "clients;";
```
Mykindos, line 41: Because some of these core tables are potentially used across multiple servers, the prefix wont work as expected.

A much larger rework will likely be required

shops/src/main/resources/plugin.yml
```diff
@@ -4,4 +4,4 @@ name: Shops
 load: POSTWORLD
 version: '1.0'
 depend: [Core]
-api-version: 1.20
\ No newline at end of file
+api-version: '1.20'
```
Mykindos, line 7: Not sure this change is needed (see other plugin.yml in the repo)

Mykindos (changes requested):
You should only need to update the readme for this

shops/src/main/resources/plugin.yml
braulio-dev: The other plugin.yml's use api-version 1.19. When Spigot loads this as a float number it will recognize it as a1.2 and not 1.20. 

This happens to anyone using 1.20 as number and not a string:
- https://www.spigotmc.org/threads/tutor-solving-unsupported-api-version-1-2-error-in-1-20-plugins.607378/
- https://stackoverflow.com/questions/76464042/error-loading-plugin-unsupported-api-version-1-2

Mykindos: All good then

README.md
```diff
@@ -1,12 +1,34 @@
...
+    ```bash
+    ./gradlew assemble
+    ```
+   
+3. Copy the generated jar files from all modules (`./build/libs/`) into your server's `plugins` folder.
+4. In your plugins folder, make sure to also include ProtocolLib, LibsDisguises, and MythicMobs as they are required dependencies.
+5. Start your server and wait for all plugins to load. You should get an error saying that the server couldn't establish a connection to the database.
+6. In the generated plugin folders, open the `config.yml` file in each one of them and fill in the required information.
+   * For `Core` specifically, fill in all database details:
+     * `prefix`: The prefix to use for all tables in this module.
```
Mykindos, line 27: prefix is filled by default, they can override it if they want to though

shops/src/main/resources/plugin.yml
Mykindos, line 7: You might as well update the others as well, maybe just do a find and replace for 

api-version: 1.19
with
api-version '1.20'

Mykindos (approved):

PR #11, opened 2023-09-21
Update Class/Skill descriptions.

StudentAlleg:
This is a pass through, updating skill descriptions to have more <val> </val> on important numbers and adding some more config options (but not as many as should be). Gives players more information on skills and makes it more readable. Future proofs changes in the config by not hardcoding some values.


StudentAlleg:
Contamination in some unintended files, fixing.

PR #17, opened 2023-09-22
Changed BuildManage GUI titles from GREEN to BLACK to improve readabi…

StudentAlleg:
…lity. Capitalized the b in <role> Builds


copytoserver.ps1
Mykindos, line 1: This should not be in source 😄 , probably put it in your start server script

Mykindos (changes requested):

copytoserver.ps1
StudentAlleg: damnit, I had already done this but it was still here on this branch and I missed it.

Mykindos (approved):

PR #20, opened 2023-09-22
vengeance ability

ezrn:


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/passives/Vengeance.java
```diff
@@ -0,0 +1,84 @@
...
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.EventPriority;
+import org.bukkit.event.Listener;
+import org.bukkit.event.entity.EntityDamageEvent.DamageCause;
+
+@Singleton
+@BPvPListener
+public class Vengeance extends Skill implements PassiveSkill, Listener {
+
+    private int numHits;
```
StudentAlleg, line 24: Num hits is never retrieved. Needs to override loadSkillConfig: `@Override
    public void loadSkillConfig(){
        numHits= getConfig("numHits", 3, Integer.class);
    }`

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/passives/Vengeance.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return "Vengeance";
+    }
+
+    @Override
+    public String[] getDescription(int level) {
+        return new String[]{
```
StudentAlleg, line 38: No usage of numHits in desciption.

StudentAlleg (changes requested):
numHits is not implemented correctly. Is this supposed to be the maximum number of hits, a tracker of hits (in which case do not retrieve it from config), or global hits (in which case it is implemented correctly)?

There does not appear to be a cap for numHits.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/passives/Vengeance.java
StudentAlleg: Does not apply if this is not the maximum hits this skill applies for.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/passives/Vengeance.java
StudentAlleg: Does not apply if this is not the maximum hits this skill applies for.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/passives/Vengeance.java
```diff
@@ -0,0 +1,89 @@
...
+                "For every subsequent hit, your damage will increase by <val>" + (level * 0.25) +"</val>",
```
StudentAlleg, line 41: Should probably have damage increase in config.

StudentAlleg (commented):
Skill increases damage per hit per player, resseting if damage is taken from any source.

StudentAlleg (approved):

Mykindos:
too similar to repeated strikes

PR #27, opened 2023-09-23
Improve all ability descriptions

ezrn:


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/axe/Flash.java
```diff
@@ -54,12 +54,12 @@ public String getName() {
     @Override
     public String[] getDescription(int level) {
         return new String[] {
-                "Teleport a short distance horizontally",
-                "in the direction you are facing.",
+                "Right click with an Axe to activate",
                 "",
-                "Uses up to <val>" + getMaxCharges(level) + "</val> charges.",
+                "Teleport 5 Blocks in the direction you are facing",
```
StudentAlleg, line 59: I think there is a teleportDistance variable that could replace 5. And also give that the <stat> tag.

Problem was I could never figure out where the 5 distance was used so I did not touch it.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/axe/Flash.java
```diff
@@ -54,12 +54,12 @@ public String getName() {
...
+                "Stores up to <val>" + getMaxCharges(level) + "</val> charges",
                 "",
-                "Gain a charge every: <stat>" + getRechargeSeconds(level) + "</stat> seconds."
+                "Gain a charge every: <stat>" + (int)getRechargeSeconds(level) + "</stat> seconds"
```
StudentAlleg, line 62: Casting to int only works if getRechargeSeconds is always guaranteed to be an int.  Which in config could be changed. _most_ times are represented as a double.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/axe/Leap.java
```diff
@@ -41,9 +41,13 @@ public String getName() {
     @Override
     public String[] getDescription(int level) {
         return new String[]{
-                "Right click with a axe to activate.",
+                "Right click with an Axe to activate",
+                "",
+                "Take a great leap forward",
+                "",
+                "Using near a wall will activate a wall-kick",
```
StudentAlleg, line 48: Is this actually how leap here is implemented? I couldn't figure it out and in testing it did not appear to work this way.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/bow/SilencingArrow.java
```diff
@@ -38,9 +38,10 @@ public String getName() {
     public String[] getDescription(int level) {
 
         return new String[]{
-                "Your next arrow will <effect>Silence</effect> your",
-                "target for <val>" + (baseDuration + level) + "</val> seconds.",
-                "Making them unable to use any active skills",
+                "Left click with a Bow to prepare",
+                "",
+                "Your next arrow will silence your", "target for <val>" + (3 + level) + "</val> seconds,",
```
StudentAlleg, line 43: Should be `(baseDuration + level)`

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/passives/RepeatedStrikes.java
StudentAlleg, line 1: casting to Int only works if that value is guaranteed to be round. This could change in config

StudentAlleg (changes requested):
Read above comments.

A general question is, do we want to use left/right click, because these keys can be rebound (for instance, we use drop instead of Q). Would be nice if you could get this info from client.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/axe/Leap.java
ezrn: yeah it has a wall kick functionality

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/axe/Flash.java
ezrn: I think it looks more consistent and better if all STATS are ints, if problems arise it can be fixed

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/axe/Flash.java
ezrn: you are right ill need to modify this to get the config

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/bow/SilencingArrow.java
ezrn: I'll redo thank you

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/axe/Flash.java
StudentAlleg: I mean, but they aren't. And you cannot guarantee it. One of the main reasons why I am putting as much into the config is so that it is much easier to change it in the future, and being able to see that update is important. Yes it looks messy, but unless we want to do logic to display a certain way (which is an option), it needs to be able to show the decimal because it _could_ be a decimal.

braulio-dev: Some statistic values need to be doubles, like cooldowns. In this specific scenario, the variable can be changed to be to a decimal number like @StudentAlleg said

ezrn: alright ill fix

PR #35, opened 2023-09-24
Swap booster and extra-damage weapons

braulio-dev:
To keep it consistent with Mineplex's game and to not confuse the rest of the audience


Mykindos (changes requested):
There is also another sql file where it maps names to these items, just search the project for "Power Sword" and you should find it

braulio-dev:
> There is also another sql file where it maps names to these items, just search the project for "Power Sword" and you should find it

Diamond swords are named Power Swords in Mineplex, I can rename radiant swords to boosters though

Mykindos (approved):

PR #39, opened 2023-09-25
Create progression system

braulio-dev:
Add in-depth progression system with four main skill trees.


core/src/main/java/me/mykindos/betterpvp/core/utilities/model/ConfigAccessor.java
Mykindos, line 1: Couldn't any class do this as long as they have access to the main JavaPlugin i.e. Progression.java

braulio-dev: I had an issue with circular dependencies I couldn't get around

Mykindos (approved):

PR #44, opened 2023-09-26
Update /c help.

StudentAlleg:
Added a page system for help command.
Added looking up a specific command.
Sorted commands by name.


StudentAlleg:
Changing to draft as there are some features I want to include, but I don't want someone else to do work I have already done.

PR #51, opened 2023-09-28
ride ability changes

ezrn:


champions/src/main/resources/config.yml
```diff
@@ -88,8 +88,7 @@ skills:
       enabled: true
       cooldown: 30
       maxlevel: 5
-      lifespan: 6.0
-      horseHealth: 5.0
```
Mykindos, line 92: not sure why you made it unconfigurable

Mykindos (changes requested):

champions/src/main/resources/config.yml
ezrn: fixed

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/sword/Ride.java
```diff
@@ -74,9 +81,9 @@ public void activate(Player player, int level) {
         horse.setStyle(Horse.Style.NONE);
         AttributeInstance horseMaxHealth = horse.getAttribute(Attribute.GENERIC_MAX_HEALTH);
         if(horseMaxHealth != null) {
-            horseMaxHealth.setBaseValue(horseHealth + ((level - 1) * 5));
+            horseMaxHealth.setBaseValue(1);
         }
-        horse.setHealth(horseHealth + ((level - 1) * 5));
+        horse.setHealth(health);
```
braulio-dev, line 86: Considering the horse should be despawned upon any damnage, instead of setting the horse's health to a value of 1, you should probably call `Entity#remove` on the horse in the `CustomDamageEvent` you're listening for above. Right now, when your horse gets hit, it dies and has you sitting in the air waiting for it to despawn and play the death animation.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/sword/Ride.java
```diff
@@ -95,6 +103,8 @@ public void run() {
                 if (!horse.isDead()) {
                     horse.remove();
                     horseData.remove(player);
+                    activeHorses.remove(horse);
+
                 }
             }
         }.runTaskLater(champions, (long) (lifespan + (level - 1)) * 20L);
```
braulio-dev, line 110: Instead of having a bunch of `BukkitRunnable`s for each horse that is spawned through this ability, you should probably setup an `@UpdateEvent` (just search for how it's used in the rest of the project), loop through the alive horses and despawn those that are either dead or invalid.

braulio-dev (changes requested):

PR #53, opened 2023-09-28
new assasisn q ability

ezrn:
Gives you speed and hearts and stuff

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/passives/Endurance.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/passives/Endurance.java
@@ -11,16 +11,21 @@ import me.mykindos.betterpvp.champions.champions.skills.types.ToggleSkill;
 import me.mykindos.betterpvp.core.components.champions.Role;
 import me.mykindos.betterpvp.core.components.champions.SkillType;
 import me.mykindos.betterpvp.core.listener.BPvPListener;
+import org.bukkit.Bukkit;
+import org.bukkit.Effect;
+import org.bukkit.Material;
+import org.bukkit.Sound;
 import org.bukkit.entity.Player;
-import org.bukkit.event.Listener;
 import org.bukkit.potion.PotionEffect;
 import org.bukkit.potion.PotionEffectType;
+import org.bukkit.attribute.Attribute;
+import org.bukkit.attribute.AttributeInstance;
 
 @Singleton
 @BPvPListener
-public class Endurance extends Skill implements ToggleSkill, CooldownSkill, Listener {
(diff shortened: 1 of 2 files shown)


StudentAlleg (changes requested):
No

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/passives/Endurance.java
```diff
@@ -0,0 +1,104 @@
...
+        AttributeInstance maxHealth = player.getAttribute(Attribute.GENERIC_MAX_HEALTH);
+        if (maxHealth != null) {
+            maxHealth.setBaseValue(26.0);
+            if ((player.getHealth()+6) > 26.0) {
+                player.setHealth(26);
+            }
+            else{
+                player.setHealth(player.getHealth()+6);
+            }
+        }
```
Mykindos, line 68: Not a good idea

Just give them their health back, instead of letting them go over standard max health.

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/passives/Endurance.java
ezrn: Thats kinda the point of the ability. Not enough testing has been done to see if its too strong or not so depending on the results of testing ill change stuff.

Mykindos: Setting the maximum health this way will almost certainly cause issues

Take a look at DefensiveAura

ezrn: I did look at defensive aura but health boost cannot give you 3 hearts only 2 and 4. I ran into no bugs while testing is there a specific thing that would break by setting the max health like this?

Mykindos: The bugs wouldn't be with the skill, but there could be scenarios where the max health gets stuck at 26 even after they dequip the set etc.

I'd be more inclined to give them 4 hearts but increase the cooldown slightly

ezrn: I did run into those issues but I fixed all the ones I could find. Ill try 4 hearts and increased CD.

Mykindos: Could do increased CD, decreased duration, lots of other variables

Its the issues you couldn't find that are the problem

PR #54, opened 2023-09-29
skill menu

RealSoy:
- made it so the book amount is the same as the skill level for easier skill building
- replaced the emerald on the top right with a gold ingot cuz it looks nicer and is the same as mineplex
- renamed repeated strikes to combo attack because mineplex had the same ability with a different name (less confusing for returning players)

[skill menu](https://ibb.co/DGKmryp)

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/builds/menus/SkillMenu.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/builds/menus/SkillMenu.java
@@ -111,7 +111,7 @@ public class SkillMenu extends Menu implements IRefreshingMenu {
         ItemStack book;
         Component name;
         if (isSkillActive(skill)) {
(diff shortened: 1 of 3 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/champions/builds/menus/SkillMenu.java
```diff
@@ -111,7 +111,7 @@ public SkillButton buildButton(Skill skill, int slot, int level) {
         ItemStack book;
         Component name;
         if (isSkillActive(skill)) {
-            book = new ItemStack(Material.ENCHANTED_BOOK);
+            book = new ItemStack(Material.WRITTEN_BOOK, (level));
```
Mykindos, line 114: dont need the parenthesis around level

Mykindos (changes requested):

Mykindos (approved):

PR #61, opened 2023-10-01
Add a warning for low energy.

StudentAlleg:
Add a title/chat warning for low energy, which is checked every 5 minutes. If your clan has less than `1.0` hours, receive a warning through title and chat.

Also fixes a bug in UtilTime.getTIme chain.

### DIFF
--- a/clans/src/main/java/me/mykindos/betterpvp/clans/clans/listeners/ClanEnergyListener.java
+++ b/clans/src/main/java/me/mykindos/betterpvp/clans/clans/listeners/ClanEnergyListener.java
@@ -38,7 +38,7 @@ public class ClanEnergyListener extends ClanListener{
         this.clans = clans;
     }
 
-    @UpdateEvent(delay = 30 * 1000, isAsync = true)
+    @UpdateEvent(delay = 300 * 1000, isAsync = true)
     public void checkEnergy() {
         Bukkit.getOnlinePlayers().forEach(player -> {
             Optional<Clan> clanOptional = clanManager.getClanByPlayer(player);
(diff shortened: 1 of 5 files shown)


clans/src/main/java/me/mykindos/betterpvp/clans/clans/listeners/ClanEnergyListener.java
```diff
@@ -0,0 +1,66 @@
...
+            }
+        });
+    }
+
+    @EventHandler(priority = EventPriority.MONITOR)
+    public void onEnergyCheck(EnergyCheckEvent event) {
+        if (event.isCancelled()) return;
+        Player player = event.getPlayer();
+        Clan clan = event.getClan();
+        if (clan.getEnergyRatio() < energyWarnLevel) {
```
Mykindos, line 57: So with the default configuration, its only going to send the titles if they are below 1 energy?

StudentAlleg: It sends it if it is below 1 hour of energy (obviously subject to change)

StudentAlleg: getEnergyRatio() gets the time in hours (I discovered)

Mykindos: I think setting the timer to every 5 minutes should be frequent enough

There will also be a /c shop command by the way, which will let clans buy energy from their base

StudentAlleg: That was the intention, was 30 seconds for testing purposes.

Could add that to the chat component when that exists too.

Mykindos (approved):

PR #62, opened 2023-10-01
added heavy arrows from mineplex (with integrated energy system)

ezrn:


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/passives/HeavyArrows.java
```diff
@@ -0,0 +1,106 @@
...
+                arrows.put(arrow, arrow.getLocation());
+
+                if (!player.isSneaking()) {
+
+                    if (championsManager.getEnergy().use(player, getName(), getEnergy(level) / 2, false)) {
+                        Vector pushback = player.getLocation().getDirection().multiply(-1);
+
+                        float charge = event.getForce();
+
+                        pushback.multiply(1.25 * charge);
```
Mykindos, line 89: Make 1.25 configurable, just a double called basePushback or something

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/passives/HeavyArrows.java
```diff
@@ -0,0 +1,115 @@
...
+import org.bukkit.util.Vector;
+
+import java.util.Iterator;
+import java.util.WeakHashMap;
+
+@Singleton
+@BPvPListener
+public class HeavyArrows extends Skill implements PassiveSkill, EnergySkill{
+
+    private final WeakHashMap<Arrow, Location> arrows = new WeakHashMap<>();
```
braulio-dev, line 31: You don't need to make this a Map if you'll only be using the key (Arrow) anyway.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/passives/HeavyArrows.java
```diff
@@ -0,0 +1,115 @@
...
+    @EventHandler
+    public void onShoot(EntityShootBowEvent event) {
+        if (!(event.getEntity() instanceof Player player)) return;
+        if (!(event.getProjectile() instanceof Arrow arrow)) return;
+
+        int level = getLevel(player);
+        if (level > 0) {
+            PlayerCanUseSkillEvent skillEvent = UtilServer.callEvent(new PlayerCanUseSkillEvent(player, this));
+            if (!skillEvent.isCancelled()) {
+                arrows.put(arrow, arrow.getLocation());
```
braulio-dev, line 85: This will make all arrows have indicator particles (including ones where you didn't have enough energy or you canceled by sneaking). You probably want to put this after the energy check

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/passives/HeavyArrows.java
```diff
@@ -0,0 +1,115 @@
...
+
+                if (!player.isSneaking()) {
+
+                    if (championsManager.getEnergy().use(player, getName(), getEnergy(level) / 2, false)) {
```
braulio-dev, line 89: What's the need to half the energy consumption here? Wouldn't this make it inconsistent with the number specified in the config

braulio-dev (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/passives/HeavyArrows.java
ezrn: fair point ill do that

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/passives/HeavyArrows.java
ezrn: to be honest I have no idea what that if statement does I just copied it from another class and it works

braulio-dev: `EnergyHandler#use(Player, String, double, boolean)` either consumes the energy you passed in (the double argument) and returns true if the player had enough energy, or returns false if the player didn't have enough energy. Here you're trying to consume x amount of energy (`getEnergy(level)`) but you're halving it, effectively making it so they only consume half the energy they're meant to. You'd just need to remove the / 2.

```suggestion
                    if (championsManager.getEnergy().use(player, getName(), getEnergy(level), false)) {
```

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/passives/HeavyArrows.java
ezrn: I don't know what that means but when I remove the map everything turns red and scary

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/passives/HeavyArrows.java
```diff
@@ -0,0 +1,119 @@
...
+            } else {
+                Player shooter = (Player) next.getShooter();
+
+                if (!shooter.isSneaking() && championsManager.getEnergy().use(shooter, getName(), getEnergy(getLevel(shooter)), false)) {
+                    Location location = next.getLocation().add(new Vector(0, 0.25, 0));
+                    Particle.CRIT_MAGIC.builder().location(location).receivers(60).extra(0).spawn();
+                }
+            }
+        }
+    }
```
Mykindos, line 78: You dont need your Map<Arrow, Location> if you replace this entire function with

```
@Override
public void displayTrail(Location location) {
        Particle.CRIT_MAGIC.builder().location(location.add(0, 0.25, 0).color(125, 0, 125).extra(0).receivers(60, true).spawn();
}
```

See another arrow skill for an example of how the trails are applied this way

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/passives/HeavyArrows.java
Mykindos, line 78: This will use energy every tick, and would stop displaying the trail half way through the shot if they stopped sneaking.

Do this stuff when the arrow is shot, that way your trail display is almost identical to longshot

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/passives/HeavyArrows.java
Mykindos: he means you can replace it with a Set

private final Set<Arrow> arrows = new HashSet<>();

ezrn: done

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/passives/HeavyArrows.java
ezrn: yeah I fixed that I just forgot to push the changes

Mykindos (commented):
I'm still not sure if this line is necessary, at this point the arrow has already been fired and is in the arrows Set.

You could leave isSneaking here, but you shouldnt need to do any energy calculations as long as you do it when the arrow is actually fired

Mykindos (approved):

PR #64, opened 2023-10-01
made life bonds user take double damage

ezrn:
-works this time


braulio-dev:
From discussion on discord:
If Life Bonds becomes too impactful on fights then they'll become the target of the enemy anyway, and introducing double damage taken to them will increase that risk by a lot for simply using the ability. If it becomes too much of an issue we can re-open this later or tune down the ability a little so it's not as rewarding as before.

Issue #73, opened 2023-10-04
/c stuck


StudentAlleg:
/c stuck: a command that players can run if they are stuck in a territory.

Issue #70, opened 2023-10-04
/c list


StudentAlleg:
Using a page like system, list all the Clans on the server. Display members online and total members, as well as specific information related to your Clan (like enemy status).

Issue #66, opened 2023-10-03
/c setenergy


StudentAlleg:
Admin command, sets the energy of your current clan to the specified number.
![csetenergy](https://github.com/Mykindos/BetterPvP/assets/36683605/53a3b00c-129b-4adb-bcb4-76164d88249d)

PR #74, opened 2023-10-04
Added the following c subcommands: setenergy, list, stuck

StudentAlleg:
1. Implements https://github.com/Mykindos/BetterPvP/issues/66. Adds setEnergy in clan.java and a new admin command /c setenergy.
2. Implements #70. Adds /c list which lists all clans in the server.
3. Implements #73. Adds /c stuck which teleports the user after a lengthy delay out of a claimed area.

### DIFF
--- a/clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
+++ b/clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
@@ -161,20 +161,13 @@ public class ClanManager extends Manager<Clan> {
         Chunk playerChunk = player.getChunk();
         World world = player.getWorld();
 
-        boolean end = false;
-
-        for (int i = 1; i < maxChunksRadiusToScan; i++) {
-            int[] offset = {-i, 0, i};
-            for (int x : offset) {
-                for (int z: offset) {
-                    Chunk chunk = world.getChunkAt(playerChunk.getX() + x, playerChunk.getZ() + z);
-                    if (getClanByChunk(chunk).isEmpty()) {
-                        chunks.add(chunk);
-                        end = true;
-                    }
+        for (int i = -maxChunksRadiusToScan; i < maxChunksRadiusToScan; i++) {
+            for (int j = -maxChunksRadiusToScan; j < maxChunksRadiusToScan; j++) {
+                Chunk chunk = world.getChunkAt(playerChunk.getX() + i, playerChunk.getZ() + j);
+                if (getClanByChunk(chunk).isEmpty()) {
+                    chunks.add(chunk);
                 }
             }
-            if (end) break;
         }
 
         if (!chunks.isEmpty()) {
@@ -182,7 +175,6 @@ public class ClanManager extends Manager<Clan> {
 
             //this should not ever happen
             if (chunk == null) return null;
-
 
             List<Location> locations = new ArrayList<>();
 
@@ -194,7 +186,9 @@ public class ClanManager extends Manager<Clan> {
             }
 
             locations.sort(Comparator.comparingInt(a -> (int) player.getLocation().distanceSquared(a)));
-            return locations.get(0);
+
+            //to prevent getting stuck in a block, add 1 to Y
+            return locations.get(0).add(0, 1, 0);
         }
         return null;
     }
--- a/clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
+++ b/clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
@@ -27,6 +27,7 @@ import net.minecraft.core.Direction;
 import org.bukkit.Bukkit;
 import org.bukkit.Chunk;
 import org.bukkit.Location;
+import org.bukkit.World;
 import org.bukkit.block.Block;
 import org.bukkit.entity.Player;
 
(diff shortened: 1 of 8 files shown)


clans/src/main/java/me/mykindos/betterpvp/clans/clans/commands/subcommands/SetEnergySubCommand.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    public String getUsage() {
+        return super.getUsage() + " <energy>";
+    }
+
+    @Override
+    public void execute(Player player, Client client, String... args) {
+
```
Mykindos, line 48: You need to check if args.length == 1

clans/src/main/java/me/mykindos/betterpvp/clans/clans/listeners/ClansMovementListener.java
```diff
@@ -123,6 +124,32 @@ public void onClanHomeTeleport(ClanHomeTeleportEvent event) {
...
+
+        Player player = event.getPlayer();
+
+        Optional<Clan> territoryOptional = clanManager.getClanByLocation(player.getLocation());
+
+        if (territoryOptional.isEmpty()) {
+            UtilMessage.message(player, "Clans", Component.text("You must be in a claimed territory to use ", NamedTextColor.GRAY)
+                    .append(Component.text("/c stuck", NamedTextColor.YELLOW)));
+            event.setCancelReason("In wilderness");
+            event.setCountdown(true);
```
Mykindos, line 139: can be replaced with

event.cancel("In wilderness");

Mykindos (changes requested):

clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
```diff
@@ -152,6 +153,37 @@ public boolean hasAccess(Player player, Location location) {
...
+                        location.add(0, 0, -i);
+                }
+                Optional<Clan> clanOptional = getClanByLocation(location);
+                if (clanOptional.isEmpty()) {
+                    locations.add(location);
+                }
+            }
+        }
+        if (!locations.isEmpty()) {
+            locations.sort(Comparator.comparingInt(a -> (int) player.getLocation().distance(a)));
```
braulio-dev, line 182: If people were to spam the `StuckSubCommand` in a claim in which they're not able to teleport out of (greater than the area you scan), this could cause performance issues not only by the sort method call but also by the distance calculation you make for each possible location, which seems to be on a per-block basis based on this implementation.

clans/src/main/java/me/mykindos/betterpvp/clans/clans/commands/subcommands/ListSubCommand.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.stream.Stream;
+
+@Singleton
+@SubCommand(ClanCommand.class)
+public class ListSubCommand extends ClanSubCommand {
+
+    @Inject
+    public ListSubCommand(ClanManager clanManager, GamerManager gamerManager) {
+        super(clanManager, gamerManager);
+        aliases.addAll(List.of("?", "h"));
```
braulio-dev, line 32: Probably want to remove these aliases since "?" and "h" are usually recognized as part of a help action.

clans/src/main/java/me/mykindos/betterpvp/clans/clans/commands/subcommands/ListSubCommand.java
```diff
@@ -0,0 +1,118 @@
...
+        if (args.length >= 1) {
+            try {
+                pageNumber = Integer.parseInt(args[0]);
+            } catch (NumberFormatException e) {
+                //pass
+            }
+        }
+
+        Collection<Clan> clansCollection = clanManager.getObjects().values();
+        List<Clan> clansList = new ArrayList<Clan>(clansCollection.stream().toList());
```
braulio-dev, line 65: You can just do
```suggestion
        List<Clan> clansList = new ArrayList<>(clanManager.getObjects().values());
```

clans/src/main/java/me/mykindos/betterpvp/clans/clans/commands/subcommands/SetEnergySubCommand.java
```diff
@@ -0,0 +1,78 @@
...
+
+        Clan playerClan = clanManager.getClanByPlayer(player).orElseThrow();
+
+        int oldEnergy = playerClan.getEnergy();
+        playerClan.setEnergy(energy);
+
+        Component component = Component.text("Energy set to: ", NamedTextColor.GRAY).append(Component.text(playerClan.getEnergy() + " - (", NamedTextColor.YELLOW)
+                .append(Component.text(playerClan.getEnergyTimeRemaining(), NamedTextColor.GOLD).append(Component.text(")", NamedTextColor.YELLOW))))
+            .append(Component.text(" Previous: ", NamedTextColor.GRAY)).append(Component.text(oldEnergy, NamedTextColor.YELLOW));
```
braulio-dev, line 65: Bit of a non-issue but in the future if you want to create `Component`s in an easier way you can just use `UtilMessage.deserialize` or plug the MiniMessage format in the `UtilMessage.message` you called below this.

braulio-dev (changes requested):

clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
Mykindos: I'll make a small change to the DelayedAction so it does nothing if the player already has one pending which will help a little.

A better solution here might be to get a list of your surrounding chunks, sort it by closest chunk, then find the best location within that chunk only to teleport the player to.

Could also look to implement a small cooldown (5s or so) on the command

clans/src/main/java/me/mykindos/betterpvp/clans/clans/commands/subcommands/ListSubCommand.java
StudentAlleg: Unintended

clans/src/main/java/me/mykindos/betterpvp/clans/clans/commands/subcommands/ListSubCommand.java
StudentAlleg: done

clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
```diff
@@ -152,6 +153,37 @@ public boolean hasAccess(Player player, Location location) {
...
+                        break;
+                    case 1:
+                        location.add(0, 0, i);
+                        break;
+                    case 2:
+                        location.add(-i, 0, 0);
+                        break;
+                    case 3:
+                        location.add(0, 0, -i);
+                }
```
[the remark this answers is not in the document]
StudentAlleg: Same height intended, as the player is teleported to the highest block. Only in the cardinal directions not intended, will be fixed by using the chunk method.

clans/src/main/java/me/mykindos/betterpvp/clans/clans/commands/subcommands/SetEnergySubCommand.java
StudentAlleg: Yeah will look into it in the future.

clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
StudentAlleg: Implemented, please check.

clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
```diff
@@ -152,6 +154,52 @@ public boolean hasAccess(Player player, Location location) {
...
+
+        Chunk playerChunk = player.getChunk();
+        World world = player.getWorld();
+
+        boolean end = false;
+
+        for (int i = 1; i < maxChunksRadiusToScan; i++) {
+            int[] offset = {-i, 0, i};
+            for (int x : offset) {
+                for (int z: offset) {
```
braulio-dev, line 169: For each `i` until `maxChunksRadiusToScan` you're always scanning the player's own chunk because one of the elements in the `offset` array is 0, meaning at one point both x and z will both be zero and pointlessly check the player's chunk.

Just replace these three for loops with two that iterate from 0 to maxChunksRadiusToScan and have one of them be x (or i) and z (or j)

Something like this:
`
for (int i = 0; i < maxChunksRadiusToScan; i++) {
    for (int j = 0; j < maxChunksRadiusToScan; j++) {
        // get the chunk (i, j)
    }
}
`

braulio-dev (changes requested):

clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
StudentAlleg: Wouldn't this still not give chunks in the negative directions relative to the player? I would think it might be better to just check if x == 0 and z == 0 and then skip.

braulio-dev: You're right, this fixes that:

```
for (int i = -maxChunksRadiusToScan; i < maxChunksRadiusToScan; i++) {
    for (int j = -maxChunksRadiusToScan; j < maxChunksRadiusToScan; j++) {
        // get the chunk (i, j)
    }
}
```

StudentAlleg: Implemented

braulio-dev (approved):

Mykindos (approved):

Issue #71, opened 2023-10-04
Allow admins to claim territory over limit.


StudentAlleg:
Allow admins in /client admin to claim more than the allowed claims. Add something to prevent overclaiming territory for Clans that are over the territory limit but have an admin in client admin (probably something that checks members to see if they are in client admin).

This is based on a specific time where something went wrong (I don't remember what, it was probably my fault) so I had to reform a Clan. There was not enough players from that Clan online, so even though I could claim past the limit, other players would come over and unclaim territory.

Combination of these two things would help fixing mistakes.

Mykindos:
They can already do this, unless you mean for a clan that wasn't created while in /client admin

StudentAlleg:
It would need to be for non-admin clans too (in case of an accidental/mistaken disband)

PR #75, opened 2023-10-05
Admin claim/unclaim QOL

StudentAlleg:
Implements #71. Admins in /client admin can claim more territory than a Clan is normally allowed to have (for handling accidental/mistaken/bugged disbands) according to their members. Territory cannot be unclaimed if the Clan that holds it has an admin in /client admin in it (to prevent unclaiming territory that would otherwise be unclaimable while in the process of restoring a Clans claims).

### DIFF
--- a/clans/src/main/java/me/mykindos/betterpvp/clans/clans/commands/subcommands/UnclaimSubCommand.java
+++ b/clans/src/main/java/me/mykindos/betterpvp/clans/clans/commands/subcommands/UnclaimSubCommand.java
@@ -8,6 +8,7 @@ import me.mykindos.betterpvp.clans.clans.commands.ClanCommand;
 import me.mykindos.betterpvp.clans.clans.commands.ClanSubCommand;
 import me.mykindos.betterpvp.clans.clans.events.ChunkUnclaimEvent;
 import me.mykindos.betterpvp.core.client.Client;
+import me.mykindos.betterpvp.core.client.ClientManager;
 import me.mykindos.betterpvp.core.command.SubCommand;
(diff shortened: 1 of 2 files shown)


clans/src/main/java/me/mykindos/betterpvp/clans/clans/commands/subcommands/UnclaimSubCommand.java
```diff
@@ -70,6 +74,15 @@ public void execute(Player player, Client client, String... args) {
                         locationClan.getName());
                 return;
             }
+
+            for (ClanMember clanMember : locationClan.getMembers()) {
+                Optional<Client> clientOptional = clientManager.getObject(clanMember.getUuid());
+                if (clientOptional.isPresent()) {
+                    if (clientOptional.get().isAdministrating()) {
+                        UtilMessage.message(player, "Clans", "You may not unclaim territory from this Clan at this time.");
```
Mykindos, line 82: you need to return here

You also dont need the clientmanager, you can use the gamermanager to get the gamer, then called gamer.getClient()

Mykindos (changes requested):

clans/src/main/java/me/mykindos/betterpvp/clans/clans/commands/subcommands/UnclaimSubCommand.java
StudentAlleg: fixed/implemented

Mykindos (approved):

PR #81, opened 2023-10-06
finally fully fixed ride

ezrn:


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/sword/Ride.java
```diff
@@ -87,22 +86,34 @@ public void activate(Player player, int level) {
...
-        new BukkitRunnable() {
-            @Override
-            public void run() {
-                if (!horse.isDead()) {
-                    horse.remove();
-                    horseData.remove(player);
-                }
+        for (Map.Entry<Player, HorseData> data : horseData.entrySet()) {
+            Player player = data.getKey();
+            int level = getLevel(player);
```
Mykindos, line 97: Instead of getting a players skill level every half second, you could also store the lifespan in HorseData

Mykindos (changes requested):

Mykindos (approved):

PR #106, opened 2023-10-11
intimidation speedrun(13:19)

ezrn:
-added this skill
-did not test


StudentAlleg:
>Did not test

Mykindos:
This is definitely way too strong, basically arctic armour without any energy cost.

Should probably stick to what Mineplex had, which was PassiveA Q ability.

I've been meaning to convert the Q skills on Assassin to be PassiveA as well

braulio-dev:
Should work now 👍

PR #127, opened 2023-10-17
fix de-blink reset orientation

ezrn:


Mykindos:
I think this makes the skill worse no cap

Mykindos (approved):

PR #132, opened 2023-10-18
redistributed power in toxic arrow

ezrn:
- Removed nausea aspect because players can turn off nausea effects in settings
- Made the arrow give Poison 2 instead of Poison 1 as compensation

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/bow/ToxicArrow.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/bow/ToxicArrow.java
@@ -80,9 +80,6 @@ public class ToxicArrow extends PrepareArrowSkill {
 
     @Override
     public void onHit(Player damager, LivingEntity target, int level) {
-        if (target.hasPotionEffect(PotionEffectType.CONFUSION)) {
-            target.removePotionEffect(PotionEffectType.CONFUSION);
-        }
         target.addPotionEffect(new PotionEffect(PotionEffectType.POISON, (int) (baseDuration + level) * 20, 1));
     }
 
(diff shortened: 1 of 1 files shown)


StudentAlleg:
Nausea is a really underwhelming effect and causes accessibility issues

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/bow/ToxicArrow.java
```diff
@@ -80,11 +80,7 @@ public Action[] getActions() {
 
     @Override
     public void onHit(Player damager, LivingEntity target, int level) {
-        if (target.hasPotionEffect(PotionEffectType.CONFUSION)) {
-            target.removePotionEffect(PotionEffectType.CONFUSION);
-        }
-        target.addPotionEffect(new PotionEffect(PotionEffectType.CONFUSION, (int) (baseDuration + level) * 20, 0));
-        target.addPotionEffect(new PotionEffect(PotionEffectType.POISON, (int) (baseDuration + level) * 20, 0));
+        target.addPotionEffect(new PotionEffect(PotionEffectType.POISON, (int) (baseDuration + level) * 20, 1));
```
Mykindos, line 83: I have a feeling this wont work the way you think it will, given how we handle poison damage.

Try give yourself poison 1 and poison 2, they will probably do the same damage

ezrn: is there a way I could modify the poison damage so that it would increase? Or perhaps I could use this as an opportunity to use "darkness" or maybe even "weakness" as those do fit the idea of a toxic arrow

Mykindos: You can leave it like this, and then create an issue so I can implement a better solution using the EffectManager

Mykindos (approved):

PR #143, opened 2023-10-21
arrow damage change

ezrn:
Arrows:
- [BUFF] Increased all arrow damage from 4.5 to 6.0

Puts arrow damage more in line with Mineplex values

Precision:
- [NERF] Decreased precision max damage to 1.0(previously 1.5)

Must be lowered to account for higher base arrow damage.

Sharpshooter:
- [NERF] Decreased sharpshooter max damage to 1.5(previously 2.25)
- [NERF] Fixed sharpshooter not resetting damage when user missed an arrow

Overall nerf to sharpshooter to account for higher base arrow damage, the damage also wasn’t properly resetting when you missed an arrow before.

Longshot:
- [NERF] Started longshot at 1 damage instead of arrow damage
- [BUFF] Buffed Longshot base damage from 18 to 23
- [BUFF] Decreased blocks-per-damage from 3.5 to 3

Longshot will now do reduced damage up close, the maximum damage cap has been increased by 5 to account for the 5 less starting damage(max damage remains the same at 29). Damage scaling has been reduced from 3.5 blocks to 3, bringing the maximum damage distance from 80.5 to 84(nerf). Overall the power of this skill has slightly lowered to account for higher base arrow damage and bring it more in line with the mineplex system.

Overcharge:
- [NERF] Made overcharge scale 1 damage per 0.5 seconds(previously 2 damage every 0.8)
- [NERF] Lowered overcharge max damage cap from 7 to 6
- [BUFF] Lowered overcharge max level from 5 to 3

Overcharge will now scale up by values of 1, fixing the previous bug that wouldn't count odd number caps. Overall the power of the skill has lowered slightly to account for higher base arrow damage.

Heavy Arrows:
- [NERF] Reduced knockback multiplier from 1.25 to 1.0

Cooked a bit too hard on these ones

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/bow/Overcharge.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/bow/Overcharge.java
@@ -9,7 +9,6 @@ import me.mykindos.betterpvp.champions.champions.skills.Skill;
 import me.mykindos.betterpvp.champions.champions.skills.data.SkillActions;
 import me.mykindos.betterpvp.champions.champions.skills.data.SkillWeapons;
 import me.mykindos.betterpvp.champions.champions.skills.types.InteractSkill;
-import me.mykindos.betterpvp.champions.champions.skills.types.PassiveSkill;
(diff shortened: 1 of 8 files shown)


Mykindos:
pretty sure I told you to stop making balancing changes

PR #157, opened 2023-10-23
added safe and unsafe subtitle displays like mineplex

ezrn:
- When you enter a safe zone, you will get a "Safe!" message in the middle of your screen
- When entering a safezone while combat tagged, a countdown will display on your screen telling you how long you are unsafe for.
- This only displays in safezones


ezrn:
- new version untested

ezrn:
Players who have dealt damage in the last 15 seconds should be treated the same as players who have taken damage in the last 15 seconds

Mykindos (approved):

PR #161, opened 2023-10-25
Proposal: Change Roles to be Configurable.

StudentAlleg:
WIP
A larger project that I was expecting, this is a proposal to change Role to be configurable, with the goal of allowing more than 6 classes and class agnostic skills. This would make changes to classes, such as armor, health values, knockback, etc. much easier.

- [x] Role Config
- [x] Skill Loading
- [x] Health calculation

Proposed structure:
It is not feasible to change the organization of skills as is, the idea being that we keep the organization of skills tied to a default class, for ease of configuration/use. These skill would still be able to be used on other classes.

Potential Problems:
Longer loading times for each class


StudentAlleg:
List of changes that affect balance due to the change.

- Backstab no longer deals 20% less damage against other assassins

StudentAlleg:
Current issues:
When changing skills for class, class still has the defaults it generated

core/src/main/java/me/mykindos/betterpvp/core/components/champions/Role.java
StudentAlleg, line 1: If someone else could take a glance and see if they see anything glaring done incorrectly I'd appreciate it.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/roles/RoleManager.java
StudentAlleg, line 1: If someone could glance through this file and see if anything is a glaring problem I would appreciate it

StudentAlleg (commented):
These two are the files with the important changes. Most things in the skill directory (which accounts for most of the files) are broad copypaste changes (like defaultClassString) and some prefix changes. The files I am asking to be glanced over are not where I think the build issue is, I just want to double check I am not commiting a coding crime that is obvious.


Code works, just all builds are linked and undefined behavior occurs when a class you have overrides are made (like a class named knight). Also generating the config is still weird. All these things are WIP. (and this is still very much WIP). But it is a proof of concept.

core/src/main/java/me/mykindos/betterpvp/core/components/champions/Role.java
Mykindos: Other than the default values being kinda weird, i.e. default armour is leather and the default skills are all knight skills (when the config is loaded and they dont exist presumably)

I can't see anything that seems 'wrong'

champions/src/main/java/me/mykindos/betterpvp/champions/champions/roles/RoleManager.java
Mykindos: This is fine

Mykindos:
This is obviously quite a massive change.

Realistically I don't see it getting merged soon (prior to release) unless it holds up incredibly well in testing.
Props for taking on such a huge refactor though

core/src/main/java/me/mykindos/betterpvp/core/components/champions/Role.java
StudentAlleg: Just wanted to make sure I was on the right track. The defaults are just there to generate the config structure, with values that go in those spots. They aren't a suggestion for a class, they are meant to be changed.

StudentAlleg:
I don't see it being done soon either, just because there are so many moving parts that will require a lot of testing. But I hope by doing this, it will make tweaking easier in the future.

StudentAlleg:
Health implemented.

StudentAlleg:
Incorrect skills are still being applied to spots (sword skills in passive spots)

StudentAlleg:
on second login after resetting database, but not closing server, sword skills are being applied to passiveA and passiveB for Assassin2

StudentAlleg:
Need to solve issues relating to champions stats, will probably wait until Mineplex Studio as it might be implementation dependent.

StudentAlleg:
Closing as not being worked on in the near future.

PR #165, opened 2023-10-27
made recall trail

ezrn:
- Now has a trail that tracks where the player was over the last x seconds
- Looks beautiful


Mykindos (changes requested):
remove any unused code

Mykindos:
Still has unused code getOldHealth

Mykindos (approved):

PR #168, opened 2023-10-27
evade revert(if desired)

ezrn:
- Makes successful evade cooldown a flat customizable internalCD again instead of based on handRaisedTime


Mykindos:
I didn't mind the scaling cooldown, I just thought the absolute minimum should be higher

PR #169, opened 2023-10-27
made rupture look epic

ezrn:
- Now takes the sounds and visuals of the blocks below it to make the rupture instead of using ice


Mykindos:
This is how we actually had it at one point, happy to do it again

Mykindos (approved):

Issue #110, opened 2023-10-11
Add Mining progression tree perks


braulio-dev:
Perks done:
- [ ] Drop multiplier
- [ ] Faster mining speed on ores/fields (?)
- [ ] XP Multiplier

StudentAlleg:
Faster mining speed not possible to do without packet modification. No method to know if a player is currently mining (for things like haste). Closest are on stop and on break. There is no way to know if they are continuing after start.

XP modifier does not make sense, enchantment xp is not used in Clans. A better idea might be to increase xp gain in fields maybe?

braulio-dev:
> Faster mining speed not possible to do without packet modification. No method to know if a player is currently mining (for things like haste). Closest are on stop and on break. There is no way to know if they are continuing after start.
> 
> XP modifier does not make sense, enchantment xp is not used in Clans. A better idea might be to increase xp gain in fields maybe?

By XP modifier I meant progression XP not vanilla experience

StudentAlleg:
> By XP modifier I meant progression XP not vanilla experience. You can also leave mining speed to me, I have an idea for it 

Makes sense. Imo it would make more sense to make the perk boost XP in fields and not elsewhere (as having it do it to all means nothing and just messes with leveling math).

StudentAlleg:
Ideas:
Have static haste in own claim/fields while holding a pickaxe

braulio-dev:
For extra mining speed if anybody wants to give it a shot before I do later this week: https://www.spigotmc.org/threads/1-8-1-13-custom-block-breaking-change-block-hardness.362586/

PR #194, opened 2023-11-04
#110 Mining Perks

StudentAlleg:
Partial implementation of #110.

- [x] Add additional drop chances for blocks that give mining experience
- [x] Give additional XP for mining in fields


clans/src/main/java/me/mykindos/betterpvp/clans/progression/perks/FieldExperienceMiningPerk.java
```diff
@@ -0,0 +1,89 @@
...
+import me.mykindos.betterpvp.progression.tree.mining.event.ProgressionMiningEvent;
+import org.bukkit.entity.Player;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+
+import java.util.function.LongUnaryOperator;
+
+@Slf4j
+@Singleton
+public class FieldExperienceMiningPerk implements Listener, ProgressionPerk {
```
braulio-dev, line 23: We already have a global fields experience multiplier, though it doesn't happen to be a perk. I'm not sure if I want to separate it into a perk, however. I kind of want to incentivize everyone to mine at fields to keep it similar to Fishing on how the main activity has to be done in fields.

progression/src/main/java/me/mykindos/betterpvp/progression/progression/perks/mining/DropMultiplierMiningPerk.java
```diff
@@ -0,0 +1,85 @@
...
+        Player player = event.getPlayer();
+        if (mining.getMiningService().getExperience(event.getBlockState().getType()) <= 0) return;
+        mining.hasPerk(player, getClass()).whenComplete((hasPerk, throwable) -> {
+            if (hasPerk) {
+                mining.getLevel(player).whenComplete((level, throwable1) -> {
+                    if (level > maxLevel) level = maxLevel;
+                    level = level - minLevel;
+                    int drops = getChance (level * increasePerLevel);
+                    if (drops == 0 || event.getItems().isEmpty()) return;
+                    event.getItems().get(0).getItemStack().add(drops - 1);
```
braulio-dev, line 73: Probably want to make sure that this doesn't go over 64. It's a very extreme case but in case we mess around with these numbers let's avoid an exception

progression/src/main/java/me/mykindos/betterpvp/progression/tree/mining/event/ProgressionMiningEvent.java
```diff
@@ -0,0 +1,26 @@
...
+@Setter
+public class ProgressionMiningEvent extends CustomEvent {
+
+    Player player;
+
+    Block block;
+
+    LongUnaryOperator experienceModifier;
+
+    public ProgressionMiningEvent(Player player, Block block, LongUnaryOperator experienceModifier) {
```
braulio-dev, line 21: I don't think we need a `ProgressionMiningEvent` if it's just a relay for `BlockBreakEvent` with an experience modifier. If its sole purpose was for changing experience we can probably come up with a different way of adding experience modifiers by maybe attaching them to the block in the mining service. However, for this case, we already handle experience for broken fields ores in [MiningListener](https://github.com/Mykindos/BetterPvP/blob/c95087de5c4f9234f88853d13b415a2a26190179/clans/src/main/java/me/mykindos/betterpvp/clans/progression/listener/MiningListener.java#L28).

progression/src/main/java/me/mykindos/betterpvp/progression/tree/mining/listener/MiningListener.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.function.LongUnaryOperator;
+
+@BPvPListener
+@Slf4j
+@Singleton
+public class MiningListener implements Listener {
+
+    @EventHandler(priority = EventPriority.MONITOR)
+    public void onBreak (BlockBreakEvent event) {
+        log.info("called");
```
braulio-dev, line 22: Debug message whoops

progression/src/main/java/me/mykindos/betterpvp/progression/tree/mining/listener/MiningListener.java
```diff
@@ -0,0 +1,31 @@
...
+        UtilServer.callEvent(new ProgressionMiningEvent(event.getPlayer(), event.getBlock(), LongUnaryOperator.identity()));
+    }
+
+    @EventHandler
+    public void onMine(ProgressionMiningEvent event) {
+        log.info("handler");
```
braulio-dev, line 28: Debug message whoops

braulio-dev (changes requested):

clans/src/main/java/me/mykindos/betterpvp/clans/progression/perks/FieldExperienceMiningPerk.java
StudentAlleg: This makes sense, this is designed to function with other experience modifiers. My next project is to add enable/disable and a standardized config for mining. I think it makes more sense as a perk.

progression/src/main/java/me/mykindos/betterpvp/progression/tree/mining/event/ProgressionMiningEvent.java
StudentAlleg: kind of makes sense, though this implementation feels easier for me to understand

progression/src/main/java/me/mykindos/betterpvp/progression/progression/perks/mining/DropMultiplierMiningPerk.java
StudentAlleg: "Adds quantity to this itemstack. Will not go over the items max stack size." it looks like the description of the add function means that it would at most be 64 even if it goes over.

progression/src/main/java/me/mykindos/betterpvp/progression/tree/mining/event/ProgressionMiningEvent.java
StudentAlleg: Removed

clans/src/main/java/me/mykindos/betterpvp/clans/progression/perks/FieldExperienceMiningPerk.java
StudentAlleg: Removed

braulio-dev (approved):

Issue #208, opened 2023-11-08
Update Config for Skills


StudentAlleg:
Give all skills as much config options as possible
All skills:

- [ ] useWhileSlowed
- [ ] useWhileSilenced
- [ ] useWhileStunned
- [ ] useWhileLevitating

- [ ] Effect Strength

Any skill that has a meaningful number, add it to config.

PR #215, opened 2023-11-09
Update Config for Skills

StudentAlleg:
#208
Add everything possible to config.


StudentAlleg:
Note: some knight axe skills have incorrect cooldown in config

StudentAlleg:
lighting orb needs to be tested

StudentAlleg (commented):
Ready for Review/Proofread, just so someone can share in the nightmare with me.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/Skill.java
```diff
@@ -73,12 +112,18 @@ public final void loadConfig() {
...
         }
 
         if (this instanceof EnergySkill) {
             energy = getConfig("energy", 0, Integer.class);
         }
+        canUseWhileSlowed = getConfig("canUseWhileSlowed", true, Boolean.class);
+        canUseWhileSilenced = getConfig("canUseWhileSilenced", false, Boolean.class);
+        canUseWhileStunned = getConfig("canUseWhileStunned", false, Boolean.class);
+        canUseWhileLevitating = getConfig("canUseWhileLevitating", false, Boolean.class);
+        canUseInLiquid = getConfig("canUseInLiquid", false, Boolean.class);
```
Mykindos, line 126: Perhaps we should only check this if the skill is NOT a passive or global skill? Might keep the config a little cleaner

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/Skill.java
StudentAlleg: Is this something that is consistent among all skills of a type? I understand the problem, I don't like how cluttered the config is.

A better solution might be to check if that config option is present for that skill, if not, use the default.

Mykindos: These limitations are only applied to Sword / Axe / Bow / Drop item skills (which are a passive I guess)

That would also be a viable solution to keep things a little cleaner

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/Skill.java
```diff
@@ -96,16 +98,36 @@ public void reload() {
...
+     * @param <T> The type of default value
+     * @return returns the config value if exists, or the default value if it does not. Does not save value in the config
+     */
+    protected <T> T getNonDefaultConfig(String name, Object defaultValue, Class<T> type) {
+        ExtendedYamlConfiguration config = champions.getConfig();
+        if (config.isSet(getPath(name))) {
+            return config.getObject(getPath(name), type);
+        }
+        return (type.cast(defaultValue));
+    }
```
Mykindos, line 128: You could put this in ExtendedYamlConfiguration (this is our own implementation)

You could likely just use config.get(key, defaultValue) that comes with the spigot config as well

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/Skill.java
StudentAlleg: config#getObject() works the best, changed to that

StudentAlleg:
Merged and solved conflicts.

Mykindos (approved):
Awesome stuff, expecting maybe a couple of bugs but its worth the trade off

PR #220, opened 2023-11-12
Fix GlowEnchant not overriding needed methods

StudentAlleg:
Solve errors being generated by GlowEnchant.


Mykindos:
These functions don't exist in the paper version specified by gradle, I would have to see the errors you are getting

PR #228, opened 2023-11-20
Level field rework(bugged)

ezrn:
- Uses damage value numbers instead of %
- Reduces damage and increases damage like mineplex
- Scales based on level
- Has a toggleable Q ability

Q ability doesn't activate for some reason
Could be something with how passiveSkill and toggleSkill interact


Mykindos:
is the Q ability not activating the same thing where it expects it to be a passive B instead of passive A or vice versa

Issue #140, opened 2023-10-20
Add enable/disable to all features


StudentAlleg:
This already exists for skills, expand to all features to allow for quick killswitching.

Features that require this:
**Clans**

- [x] Insurance
- [x] Pillage
- [x] Dominance

**Core**

- [x] Tips

**Progression**

- [x] Fishing Tree
- [x] Mining Tree
- [x] Fishing bait
- [x] Fishing Loot

Mykindos:
Do you have a list of features that doesn't have it, I think most do

StudentAlleg:
No, I know a lot do have it. Specifically know that progression perks do not currently.

Mykindos:
@braulio-dev 😠

StudentAlleg:
I'll put going through everything and making sure it exists on my to-do list

braulio-dev:
Do you mean just a switch for individual perks?

StudentAlleg:
Yes for perks, I don't know if it exists for trees. This should be something I am capable of doing, after my Fishing and Fishing-Plus PRs.

StudentAlleg:
@braulio-dev @Mykindos updated original post with features that do not have it.

PR #229, opened 2023-11-20
Add basic enable/disable to progression and skills

StudentAlleg:
Partial implementation of #140, basic enable/disable for progression


clans/src/main/java/me/mykindos/betterpvp/clans/progression/perks/BaseFishingPerk.java
```diff
@@ -88,5 +92,6 @@ public void onFish(PlayerStartFishingEvent event) {
     @Override
     public void loadConfig(@NotNull ExtendedYamlConfiguration config) {
         this.requiredLevel = config.getOrSaveObject("fishing.base-fishing-perk.level", 500, Integer.class);
+        this.enabled = config.getOrSaveBoolean("fishing.base-fishing-perk.enabled", true);
```
Mykindos, line 95: Is there any particular reason why you did this one via loadConfig, and then the rest via the config injector?

StudentAlleg: This one was done by Rey via ConfigAccessor, which I completely missed until I put it in here. I probably should do everything the same way, it makes way more sense to do it via ConfigAccessor.

Mykindos (approved):

PR #244, opened 2023-12-16
Add Blood Compass

ezrn:
- Tested with one target, needs testing with multiple
- Passive A bug needs to be fixed, and then this ability needs to be moved to passive A

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/passives/BloodCompass.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/passives/BloodCompass.java
@@ -10,6 +10,7 @@ import me.mykindos.betterpvp.champions.champions.skills.types.ToggleSkill;
 import me.mykindos.betterpvp.core.components.champions.Role;
 import me.mykindos.betterpvp.core.components.champions.SkillType;
 import me.mykindos.betterpvp.core.listener.BPvPListener;
+import me.mykindos.betterpvp.core.utilities.UtilBlock;
 import me.mykindos.betterpvp.core.utilities.UtilMessage;
 import me.mykindos.betterpvp.core.utilities.UtilPlayer;
 import org.bukkit.*;
@@ -54,7 +55,7 @@ public class BloodCompass extends Skill implements ToggleSkill, CooldownSkill {
                 "Shoot out up to <val>" + (numLines + (level-1)) + "</val> tracking lines that will fly",
                 "towards the nearest enemies within <val>" + (maxDistance * level) +"</val> blocks",
                 "",
-                "Players hit with these blood lines will recieve <effect>Glowing</effect> and <effect>Darkness</effect> for <stat>" + effectDuration + "</stat> seconds",
+                "Players hit with these blood lines will take <stat>7<stat> damage and recieve <effect>Glowing</effect> for <stat>" + effectDuration + "</stat> seconds",
         };
     }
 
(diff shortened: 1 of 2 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/passives/BloodCompass.java
```diff
@@ -0,0 +1,267 @@
...
+    public String getName() {
+        return "Blood Compass";
+    }
+
+    @Override
+    public String[] getDescription(int level) {
+        return new String[]{
+                "Drop Sword / Axe to activate",
+                "",
+                "Shoot out up to <val>" + (numLines + (level-1)) + "</val> tracking lines that will fly",
```
StudentAlleg, line 54: Calculations should probably be in their own functions, so that when something is changed in the skill, it is reflected in the description and vice versa.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/passives/BloodCompass.java
```diff
@@ -0,0 +1,267 @@
...
+                "towards the nearest enemies within <val>" + (maxDistance * level) +"</val> blocks",
+                "",
+                "Players hit with these blood lines will recieve <effect>Glowing</effect> for <stat>" + effectDuration + "</stat> seconds",
+                "and take <val>" + damage + ((level-1) * 2) +"</val> damage",
```
StudentAlleg, line 58: Anything to do with level (and even things that don't have to do with level) should have a modifier per level. See https://github.com/Mykindos/BetterPvP/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/axe/Cleanse.java#L64. This makes it really easy to change how a level should interact with the skill.

StudentAlleg (changes requested):
Needs some changes to conform with other skills in terms of config use and how changed values are handled in descriptions and implementation.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/passives/BloodCompass.java
```diff
@@ -0,0 +1,320 @@
...
+                "Cooldown: <val>" + getCooldown(level)
+        };
+    }
+
+    public double getFinalDamage(int level){
+        return (damage + ((level - 1)) * damageIncreasePerLevel);
+    }
+
+    public int getFinalNumLines(int level){
+        return (numLines + (level - 1));
```
StudentAlleg, line 71: It would be nice to have an IncreasePerLevel here, though that might just be me being pedantic.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/passives/BloodCompass.java
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    private void findEnemies(Player player, int level) {
+        markers.clear();
+        lineToPlayerMap.clear();
+        List<Player> enemies = UtilPlayer.getNearbyEnemies(player, player.getLocation(), (maxDistance));
+        enemies.sort(Comparator.comparingDouble(p -> p.getLocation().distance(player.getLocation())));
+        enemies = enemies.subList(0, Math.min(enemies.size(), numLines * level));
```
StudentAlleg, line 79: Does not use getFinalNumLines

StudentAlleg (changes requested):
It looks like maxDistance and escapeRadius at least should have an IncreasePerLevel (even if that increase is 0). Only needed change from my end is to use getFinalLineCount somewhere other than in description.

Mykindos (approved):

PR #246, opened 2023-12-16
made flesh hook function similarly to mineplex

ezrn:
- Now only shoot one hook instead of 5
- Now does damage which scales based on level and the amount the hook was charged
- Everything else remains the same


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/sword/FleshHook.java
```diff
@@ -66,6 +71,7 @@ public String[] getDescription(int level) {
                 "Hold right click with a Sword to channel",
                 "",
                 "Charge a hook that latches onto enemies, pulling them towards you",
+                "and dealing up to <val>" + (damage + (damageIncreasePerLevel * (level-1)))+"</val> damage",
```
StudentAlleg, line 74: This should be in a separate function just for ease of maintaining the file.

core/src/main/java/me/mykindos/betterpvp/core/combat/throwables/ThrowableItem.java
```diff
@@ -43,6 +45,14 @@ public ThrowableItem(Item item, LivingEntity thrower, String name, long expireTi
         immune = new ArrayList<>();
     }
 
+    public void setChargePercent(int chargePercent) {
```
StudentAlleg, line 48: I don't understand why you are doing this here, and not keeping it contained in FleshHook. Whenever you shoot a fleshhook, you should be able to keep track of it in a WeakHashmap, and on checking for collision you check to see if that fleshhook is in your list.

StudentAlleg (changes requested):

Mykindos (approved):

Issue #209, opened 2023-11-09
Skills that rely on canHurt may end up benefiting enemies


Mykindos:
Skills that provide allies buffs that check the relationship based on canHurt may end up buffing enemies if they are in a safe zone

Function to fetch nearby enemies / allies may need to be changed

LOW PRIORITY

StudentAlleg:
FetchNearbyPropertyEvent is what populates the EntityProperty, which is done using canHurt. So all abilities that affect allies apply to all players that that player cannot hurt. Change would be to check for Clan in Clans (which iirc is done in the clansSkillListener)/ally instead of whether damage is allowed.

Mykindos:
Fixed by https://github.com/Mykindos/BetterPvP/pull/250 I believe

PR #250, opened 2023-12-18
Change how onFetchNearbyEntity populates

StudentAlleg:
#209 Changed checking canHurt to new function, isAlly.


Mykindos:
What if a player is in a safe zone, but not your clans ally. The canHurt would return false in this scenario

StudentAlleg:
> What if a player is in a safe zone, but not your clans ally. The canHurt would return false in this scenario

Nothing was changed about canHurt, isAlly is only used to check relationship. CanHurt would return false, and prevent any damage https://github.com/Mykindos/BetterPvP/blob/262e04cfbb7144d0ffce2ed736e844d8e69257b7/clans/src/main/java/me/mykindos/betterpvp/clans/clans/listeners/ClansCombatListener.java#L60

Mykindos (approved):

Issue #206, opened 2023-11-08
Prevent creating a clan with the same name as a command


Mykindos:
i.e. /c create disband /c create ally, etc should not be permitted

StudentAlleg:
I kinda have an inkling of how to do this. I'd also assume that we'd be using Mineplex's isChatFiltered to check Clannames in general.

Mykindos:
> I kinda have an inkling of how to do this. I'd also assume that we'd be using Mineplex's isChatFiltered to check Clannames in general.

Yeah, wouldn't want some some gamers making edgy clan names.
It could probably be cancelled in the CreateClanEvent by using the Clans injector to get an instanceof ClanCommand, then looping through all subcommands

Mykindos:
Fixed in https://github.com/Mykindos/BetterPvP/pull/251

PR #251, opened 2023-12-18
Prevent clan's subcommands from being a clanname

StudentAlleg:
#206


clans/src/main/java/me/mykindos/betterpvp/clans/clans/listeners/ClanEventListener.java
```diff
@@ -62,17 +46,20 @@ public class ClanEventListener extends ClanListener {
     private final WorldBlockHandler blockHandler;
     private final Clans clans;
 
+    private final ClanCommand clanCommand;
```
Mykindos, line 49: While this technically is more performant, it seems wrong inject this particular instance

I'd recommend injecting the CommandManager, and using getCommandByAlias to store or access the ClanCommand

Mykindos (changes requested):

clans/src/main/java/me/mykindos/betterpvp/clans/clans/listeners/ClanEventListener.java
StudentAlleg: implemented

Mykindos (approved):

Issue #253, opened 2023-12-18
Make loading default builds safe.


StudentAlleg:
If a default skill is disabled, points are still taken.

Issue #210, opened 2023-11-09
Disabled skills dont refund skill points


Mykindos:
When disabling a skill in config, players may end up stuck with a certain skill equipped that they can not take any points out of as the skill gets hidden from the build menu.

A work-around for this is to just right click any skill of the same type, and it will remove the points.
Would probably be better to just refund the points next time they open the build menu if the current skill is disabled

Mykindos:
Fixed in https://github.com/Mykindos/BetterPvP/pull/254

PR #254, opened 2023-12-18
Update build loading.

StudentAlleg:
#253, #210. On champions reload, disabled skills are now removed from builds and points refunded. Default builds now use the same method as saved builds, and won't take points from disabled/null skills. A new notification is given to players if a skill they had is no longer available.

This does not update builds in the database, if a player does not change their build in the build table, it will continue to give the warning.

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/builds/BuildManager.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/builds/BuildManager.java
@@ -3,8 +3,13 @@ package me.mykindos.betterpvp.champions.champions.builds;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import lombok.Getter;
+import me.mykindos.betterpvp.champions.Champions;
+import me.mykindos.betterpvp.champions.champions.builds.event.LoadBuildsEvent;
 import me.mykindos.betterpvp.champions.champions.builds.repository.BuildRepository;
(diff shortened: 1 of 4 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/champions/builds/BuildManager.java
```diff
@@ -3,17 +3,44 @@
...
+        getBuildRepository().loadDefaultBuilds(builds);
+        addObject(player.getUniqueId().toString(), builds);
+        UtilServer.runTask(champions, () -> {
+            UtilServer.callEvent(new LoadBuildsEvent(player, builds));
+        });
     }
+
+    public void reloadBuilds() {
+        getObjects().clear();
+        for (Player player : Bukkit.getOnlinePlayers()) {
```
Mykindos, line 40: What about players who are offline at the time of reload

StudentAlleg: Whenever players logout, their data is removed. On reload, it just runs the same thing that happens for players on login, for every player. Players that have disabled skills in their builds will need to update their builds. If they do not, they will receive a message next time they login, as disabling a skill does not remove it from the database.

For #210 the issue was never offline players, as loading builds on login correctly did not include disabled skills (and did not take the points), with the exception of generating new default builds if a default skill was disabled.

Mykindos (approved):

Mykindos (approved):

Issue #160, opened 2023-10-25
Add recipe system to create / block recipes


Mykindos:

StudentAlleg:
https://jd.papermc.io/paper/1.20/org/bukkit/inventory/ShapedRecipe.html seems like the best approach, along with a config for enabling the craft.

StudentAlleg:
Current approach I am working on is using https://jd.papermc.io/paper/1.20/org/bukkit/Bukkit.html#recipeIterator() to iterate through recipes, and removing recipes that are disabled.

PR #256, opened 2023-12-20
Custom Item Overhaul

StudentAlleg:
#160 - Allow for the easy(ish) creation of crafting recipes, and a system to prevent unwanted recipes.

- [x] Block specified recipes from config
- [x] Easy implementation of recipes in code
- [x] Items are loaded from databases
- [x] Items that need code components load from database, and replace items in itemHandler
- [x] Custom Durability


StudentAlleg (commented):
Ready for review. I am planning on going over this in the coming days, as this has felt semi-spaghetti, but I first need to get a break from this specific code and come back to it with fresh eyes.

Features Include:

- Ability to disable recipes through config (crafting, smelting, anything that is a keyed recipe) 
- Ability to spawn custom items
- Reworked internal item handling, now goes by Namedspacedkey#asString()
- Added custom durability, items break after 0
- Standard item fields are defined in database, not in code
- Items that require a code component load standard item fields on runtime
- Custom/Overridden items consist of a Material +  CustomModelData. A custom item is defined by these things, all other aspects can be applied at different times.

core/src/main/java/me/mykindos/betterpvp/core/command/commands/admin/CustomGiveCommand.java
```diff
@@ -0,0 +1,83 @@
...
+        /*if (args.length > 3) {
+            try {
+                count = Integer.parseInt(args[2]);
+                if (count < 1) {
+                    throw new NumberFormatException("Number must be greater than or equal to 1");
+                }
+            } catch (NumberFormatException e) {
+                UtilMessage.message(player, UtilMessage.deserialize("<green>%s>/green> is not a valid number. Accepts integers greater than 0", args[2]));
+            }
+        }*/
```
Mykindos, line 62: remove?

core/src/main/java/me/mykindos/betterpvp/core/items/BPVPItem.java
```diff
@@ -1,24 +1,288 @@
...
+    }
+
+    /**
+     * Damage an item of this type
+     * @param player the player damaging
+     * @param itemStack the ItemStack to damage
+     * @param damage the damage the ItemStack should take
+     * @return the damaged ItemStack
+     */
+    public ItemStack damageItem (Player player, ItemStack itemStack, int damage) {
```
Mykindos, line 222: remove space after damageItem (

Mykindos (changes requested):

Mykindos (changes requested):
Some things from a quick test:

- Log messages sent when breaking a block can be removed (probably sends it in other situations as well)
- Damage to armour should probably just be kept to 1, instead of the damage dealt. Keeps it in line with weapons and probably easier to balance later on.
- There is quite a few possible null pointer warnings around usage of dataContainer.get(CoreNamespaceKeys.DURABILITY_KEY, PersistentDataType.INTEGER)

    some of these seem to just be IDE having a spazm and likely would never cause issues, but warnings are an eye sore. I'd 
    recommend in some cases just doing get() and then checking if it is null instead of calling has
- I'd like to see if there is something we can do with the actual item durability to make it visible without needing to hover the item. Perhaps setting it to a percentage depending on the artificial durability. There may also be a way to change it without messing with the items vanilla durability at all. I don't expect you to do this as part of your PR, its something I would happily look into afterwards

StudentAlleg:
> Some things from a quick test:
> 
>     * Log messages sent when breaking a block can be removed (probably sends it in other situations as well)
> 
>     * Damage to armour should probably just be kept to 1, instead of the damage dealt. Keeps it in line with weapons and probably easier to balance later on.
> 
>     * There is quite a few possible null pointer warnings around usage of dataContainer.get(CoreNamespaceKeys.DURABILITY_KEY, PersistentDataType.INTEGER)
>       some of these seem to just be IDE having a spazm and likely would never cause issues, but warnings are an eye sore. I'd
>       recommend in some cases just doing get() and then checking if it is null instead of calling has
> 
>     * I'd like to see if there is something we can do with the actual item durability to make it visible without needing to hover the item. Perhaps setting it to a percentage depending on the artificial durability. There may also be a way to change it without messing with the items vanilla durability at all. I don't expect you to do this as part of your PR, its something I would happily look into afterwards

1. Fixed
2. Changed, makes sense
3. It shouldn't ever get a null pointer because it is always checked, but I think I am going to move some of those functions into UtilItem
4. Yeah, I was thinking about that. Probably will end up setting damage based on the ration of custom durability/maxDurability. But could be confusing with the numbers, wonder if that is hideable

Mykindos (approved):

PR #277, opened 2023-12-29
Remove extra tick extension on blocking

braulio-dev:
Changes:
- Migrate to state-based checking for sword and shield right clicking.
- Remove extra tick extension on all right-clicks for every other item.

Changes were tested.


braulio-dev:
Changes may have unintended behaviors on server instances without the private blocking module. Looking into it

braulio-dev:
Solution was to save time since last block instead of blocking state and implement a grace period for cosmetic shields (used in default behavior implementation) before the server is able to register the player is blocking.

Mykindos (approved):

PR #279, opened 2023-12-29
Remove hard redis requirement to host a BetterPvP server

braulio-dev:
Changes:
- Add `enabled` boolean option in Core `config.yml` under the `redis` subsection.
- `password` field in the `redis` subsection located in the Core `config.yml` is no longer required.
- Added logging for Redis enable/disable.

Changes were tested.

### DIFF
--- a/core/src/main/java/me/mykindos/betterpvp/core/redis/Redis.java
+++ b/core/src/main/java/me/mykindos/betterpvp/core/redis/Redis.java
@@ -1,19 +1,54 @@
 package me.mykindos.betterpvp.core.redis;
 
+import com.google.common.base.Preconditions;
 import com.google.inject.Singleton;
+import lombok.extern.slf4j.Slf4j;
 import org.bukkit.configuration.ConfigurationSection;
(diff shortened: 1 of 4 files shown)


core/src/main/java/me/mykindos/betterpvp/core/redis/Redis.java
```diff
@@ -1,19 +1,54 @@
...
+
+        final JedisPoolConfig poolConfig = getJedisPoolConfig();
+        if (redisCredentials.getPassword() == null || redisCredentials.getPassword().isEmpty()) {
+            // Passwordless redis
+            this.pool = new JedisPool(
+                    poolConfig,
+                    redisCredentials.getHost(),
+                    redisCredentials.getPort(),
+                    0,
+                    null,
```
Mykindos, line 32: i dont think this will actually make any difference because config.getString("password") was nullable, you can even see that you are checking if redisCredentials.getPassword() == null, so passing that in would be same as passing null

braulio-dev: good catch

PR #289, opened 2024-01-04
Curse Warlock Passive Draft 

ezrn:
- Hits give enemies Vulnerability
- Internal cooldown for when Vulnerability can be applied
- If user takes damage Curse effect is removed from all targets


ezrn:
bad ability

PR #304, opened 2024-01-07
winblade sound and backstab sound

RealSoy:
- changed windblade flying sound to mineplex's because the one now is ear piercingly loud and there is no way to turn it down in the minecraft sound settings
- added mineplex's tick sound to backstab because it's so satisfying

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/passives/Backstab.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/passives/Backstab.java
@@ -12,6 +12,7 @@ import me.mykindos.betterpvp.core.components.champions.SkillType;
 import me.mykindos.betterpvp.core.listener.BPvPListener;
 import me.mykindos.betterpvp.core.utilities.UtilMath;
 import me.mykindos.betterpvp.core.utilities.UtilPlayer;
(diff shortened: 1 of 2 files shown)


ezrn:
import sound and it will work

Mykindos:
Did you actually test this? Because that definitely aint the sound

Old sound was Fizz, which maps directly to FIZZ("FIZZ", "BLOCK_FIRE_EXTINGUISH")
block.fire.extinguish is identical to block.lava.extinguish

RealSoy:
it sounded the exact same to me,  ill check the osund on soyplex

RealSoy:
i checked and im pretty sure the fizz sound is the sound that plays if u fly on the ground for too long

https://streamable.com/h5dlvc bpvp
https://streamable.com/2eijh5 mineplex

braulio-dev:
> i checked and im pretty sure the fizz sound is the sound that plays if u fly on the ground for too long
> 
> https://streamable.com/h5dlvc bpvp https://streamable.com/2eijh5 mineplex

You have a texturepack on, that's why BetterPvP windblade sounds like that. I sort of prefer the newer sound we have since it gives a clearer indication that somebody is using the windblade around you instead of a stealthy cue.

ezrn:
I've added the backstab sound to one of my other PRs with the change to Mineplex's functionality

PR #317, opened 2024-01-12
Vault Blocks

ezrn:
NOT YET TESTED


ezrn:
seeing as the raiding system is going to cannon raiding this is now irrelevant

Issue #296, opened 2024-01-06
Dropping throwing web triggers cooldown. Cooldown does not stop throwing more webs


StudentAlleg:
Dropping throwing web triggers cooldown. Does not throw web. Cooldown does not stop throwing more webs (in general)

braulio-dev:
Happens with any interactable action in the game. For example:
- Skills
- Energy Apples
- Fishing Baits

StudentAlleg:
working specifically on cooldown for web, moving other to new issue

StudentAlleg:
Cooldown is only effective when using item, not when attacking with item

PR #352, opened 2024-01-13
Fix left click not giving cooldown for throwing_web

StudentAlleg:
#296

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/weapons/weapons/ThrowingWeb.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/weapons/weapons/ThrowingWeb.java
@@ -10,12 +10,14 @@ import me.mykindos.betterpvp.core.combat.throwables.ThrowableItem;
 import me.mykindos.betterpvp.core.combat.throwables.events.ThrowableHitEntityEvent;
 import me.mykindos.betterpvp.core.combat.throwables.events.ThrowableHitGroundEvent;
 import me.mykindos.betterpvp.core.config.Config;
+import me.mykindos.betterpvp.core.cooldowns.CooldownManager;
 import me.mykindos.betterpvp.core.framework.CoreNamespaceKeys;
 import me.mykindos.betterpvp.core.items.BPVPItem;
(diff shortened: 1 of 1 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/weapons/weapons/ThrowingWeb.java
```diff
@@ -85,7 +90,10 @@ public void activate(Player player) {
     @EventHandler
     public void onInteract(PlayerInteractEvent event) {
         if (event.getAction().isLeftClick() && matches(event.getItem())) {
-            activate(event.getPlayer()); // also activate on left click
+            String name = PlainTextComponentSerializer.plainText().serialize(getName());
```
braulio-dev, line 93: You should probably use `BPVPItem#getKey` instead of `BPVPItem#getName` since the former one has to be unique for all items and is already a string, whereas the latter one is an adventure component, and for classes that use the default constructor is the same component with content `Unintialized Weapon`, meaning cooldown would be shared across those weapons

StudentAlleg: This is using the same thing as where the normal cooldown is being applied. I think it makes more sense to refactor both (it makes more sense). But getName would be overwritten when the item is loaded, so not an issue there.

braulio-dev: Still should probably use the key anyway, better practice than to use the display name

StudentAlleg: the display name is also what is used to display to the player, which key was not really intended to

braulio-dev: Oh I see what you mean, the cooldown manager displays that string. All good then

braulio-dev (approved):

PR #362, opened 2024-01-13
bleed effect, sever and hilt smash fixes/changes

ezrn:


Mykindos:
doesnt build

ezrn:
i broke it all

PR #374, opened 2024-01-17
Fix/misc bugs 12.1.24

### LINKED ISSUE #338: Smoke bomb does not re-appear while breaking blocks or swing hand
For smokebomb, I'd kind of like to see a longer duration, but can't attack while invisible (must reappear first, maybe have a slight cooldown)

### LINKED ISSUE #341: Magma Blade is not giving extra damage

### LINKED ISSUE #342: /kick does not work with a mutli-word reason

### LINKED ISSUE #346: dying in combat sometimes makes you unsafe when you respawn

### LINKED ISSUE #348: add damage report when you die/get a kill which you can hover to see sources of damage

### LINKED ISSUE #349: book is bugged
does not give correct values

### LINKED ISSUE #350: Dropping item triggers cooldown.
Happens with any interactable action in the game. For example:
- Skills
- Energy Apples
- Fishing Baits

This is likely due to an update in higher versions forcing the player to swing their arm when an item is dropped and the server takes that swing as an interaction.

_Originally posted by @braulio-dev in https://github.com/Mykindos/BetterPvP/issues/296#issuecomment-1884273088_
            
Happens with throwing webs, energy apple

### LINKED ISSUE #359: Wolfpounce bug
I know the ability is being changed to chargedata but if you block in the air and then touch the ground it should start charging rather than blocking while jumping and the ability doesnt active unless you block again

### LINKED ISSUE #360: Skill Token Bug
Hard to explain but heres an example:
if you have 3 tokens in long shot it the other passive As display values as if they were also level 3

Also, if you have no skill tokens it should display as (0/X) value next to it to clarify that the values are the baseline

braulio-dev:
Fixes #338, fixes #341, fixes #342, fixes #346, fixes #348, fixes #348, fixes #349, fixes #350, fixes #359, fixes #360

Changes were tested

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/passives/SmokeBomb.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/passives/SmokeBomb.java
@@ -17,6 +17,7 @@ import me.mykindos.betterpvp.core.utilities.UtilMessage;
 import me.mykindos.betterpvp.core.utilities.UtilPlayer;
 import me.mykindos.betterpvp.core.utilities.UtilServer;
 import me.mykindos.betterpvp.core.utilities.UtilTime;
+import org.bukkit.Bukkit;
 import org.bukkit.Particle;
 import org.bukkit.Sound;
 import org.bukkit.entity.Player;
@@ -30,14 +31,16 @@ import org.bukkit.potion.PotionEffectType;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import java.util.HashMap;
 import java.util.Iterator;
-import java.util.WeakHashMap;
+import java.util.Map;
+import java.util.UUID;
 
 @Singleton
 @BPvPListener
 public class SmokeBomb extends Skill implements ToggleSkill, CooldownSkill, Listener {
 
-    private final WeakHashMap<Player, Long> smoked = new WeakHashMap<>();
+    private final Map<UUID, Long> smoked = new HashMap<>();
 
     private double baseDuration;
     private double durationIncreasePerLevel;
@@ -101,7 +104,7 @@ public class SmokeBomb extends Skill implements ToggleSkill, CooldownSkill, List
     public void toggle(Player player, int level) {
         // Effects
         championsManager.getEffects().addEffect(player, EffectType.INVISIBILITY, (long) (getDuration(level) * 1000L));
-        smoked.put(player, System.currentTimeMillis());
+        smoked.put(player.getUniqueId(), System.currentTimeMillis());
         for (Player target : UtilPlayer.getNearbyEnemies(player, player.getLocation(), blindRadius)) {
             target.addPotionEffect(new PotionEffect(PotionEffectType.BLINDNESS, (int) (blindDuration * 20), 0));
         }
@@ -132,19 +135,19 @@ public class SmokeBomb extends Skill implements ToggleSkill, CooldownSkill, List
     }
 
     private void interact(Player player) {
-        final long castTime = smoked.get(player);
+        final long castTime = smoked.get(player.getUniqueId());
         if (!UtilTime.elapsed(castTime, 25L)) {
             return;
         }
 
-        smoked.remove(player);
+        smoked.remove(player.getUniqueId());
         reappear(player);
     }
 
     @EventHandler
     public void onInteract(PlayerInteractEvent event) {
         Player player = event.getPlayer();
-        if (smoked.containsKey(player)) {
+        if (smoked.containsKey(player.getUniqueId())) {
             interact(player);
         }
     }
@@ -152,7 +155,7 @@ public class SmokeBomb extends Skill implements ToggleSkill, CooldownSkill, List
     @EventHandler
     public void onPickup(PlayerAttemptPickupItemEvent event) {
         Player player = event.getPlayer();
-        if (smoked.containsKey(player)) {
+        if (smoked.containsKey(player.getUniqueId())) {
             interact(player);
         }
     }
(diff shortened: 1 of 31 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/passives/SmokeBomb.java
```diff
@@ -10,162 +10,62 @@
...
 import java.util.Iterator;
-import java.util.Map.Entry;
-import java.util.Optional;
 import java.util.WeakHashMap;
 
 @Singleton
 @BPvPListener
 public class SmokeBomb extends Skill implements ToggleSkill, CooldownSkill, Listener {
 
+    private final WeakHashMap<Player, Long> smoked = new WeakHashMap<>();
```
Mykindos, line 40: I still don't like using players in a weakhashmap, you can't listen to when an object is gc'd and it may prevent clean up in other places, im also gonna go ahead and assume this is more overhead compared to a standard hashmap

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/types/CooldownSkill.java
```diff
@@ -18,12 +17,7 @@ default boolean isCancellable(){
     }
 
     default boolean shouldDisplayActionBar(Gamer gamer) {
-        final Player found = gamer.getPlayer();
-        if (found == null || !hasSkill(found)) {
-            return false;
-        }
-
-        return SkillWeapons.getTypeFrom(found.getInventory().getItemInMainHand()) == getType();
+        return isHolding(gamer.getPlayer()) && (getType() == SkillType.AXE || getType() == SkillType.SWORD);
```
Mykindos, line 20: Is this also the cool down action bar? What about drop item skills such as smoke bomb

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/sword/WolfsPounce.java
```diff
@@ -171,40 +140,51 @@ public void loadSkillConfig() {
...
+        // Action bar
+        Gamer gamer = championsManager.getClientManager().search().online(player).getGamer();
+        gamer.getActionBar().add(900, actionBarComponent);
     }
 
     @Override
-    public boolean shouldDisplayActionBar(Gamer gamer) {
-        return false;
+    public void invalidatePlayer(Player player) {
+        // Action bar
```
Mykindos, line 151: I know this is unchanged, but wouldn't it be wiser if we just pass the gamer and player as arguments into invalidatePlayer and similar? I imagine this code to get the gamer is quite frequent?

If we only do it a couple of times don't worry about it

core/src/main/java/me/mykindos/betterpvp/core/combat/click/RightClickListener.java
```diff
@@ -37,13 +38,28 @@ public class RightClickListener implements Listener {
...
     public RightClickListener(ClientManager clientManager, Core core) {
         this.clientManager = clientManager;
         this.core = core;
     }
 
+    // Fix for interact event triggering when dropping items
+    @EventHandler
+    public void onDrop(PlayerDropItemEvent event) {
+        lastDrop.put(event.getPlayer(), System.currentTimeMillis());
+    }
```
Mykindos, line 53: Is this necessary still? Or is this part of the eventresult.deny stuff

core/src/main/java/me/mykindos/betterpvp/core/combat/damagelog/DamageLogListener.java
```diff
@@ -4,18 +4,25 @@
...
 @BPvPListener
 public class DamageLogListener implements Listener {
 
-    private final DamageLogManager damageLogManager;
-
     @Inject
-    public DamageLogListener(DamageLogManager damageLogManager) {
-        this.damageLogManager = damageLogManager;
-    }
+    private DamageLogManager damageLogManager;
```
Mykindos, line 25: Why?

core/src/main/java/me/mykindos/betterpvp/core/combat/events/CustomDamageReductionEvent.java
```diff
@@ -1,15 +1,20 @@
...
 public class CustomDamageReductionEvent extends CustomEvent {
 
     private final CustomDamageEvent customDamageEvent;
+    private final double initialDamage;
     private double damage;
+
+    public CustomDamageReductionEvent(CustomDamageEvent customDamageEvent, double damage) {
+        this.customDamageEvent = customDamageEvent;
+        this.damage = damage;
+        this.initialDamage = damage;
```
Mykindos, line 18: Can't you just get this from the customdamageevent

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/types/CooldownSkill.java
braulio-dev: We had those on but they were conflicting with other skill's action bar, I think that's also the reason Mineplex removed it.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/passives/SmokeBomb.java
braulio-dev: I don't see the first two problems as an issue because in the case of skills, it's not really important to us if we know when an object is GC'd or not, and weak maps store only weak references to its keys so it won't prevent cleanup elsewhere. I don't really know if it's less performant and resource-consuming but I would assume it's better than us having to manually clean entries up by calling `Bukkit.getPlayer` to check if they're still online.

I believe the getPlayer variation with a UUID parameter is a constant execution time as well so I'll go ahead and change the map to use UUID keys. We could probably do this for most skills as well.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/sword/WolfsPounce.java
braulio-dev: We do it exactly 10 times in that same method, I suppose we can do that, yes.

core/src/main/java/me/mykindos/betterpvp/core/combat/click/RightClickListener.java
braulio-dev: Yeah it's part of that bugfix. I think an alternative is to mess with server internals but that's probably a lot more work than we think

core/src/main/java/me/mykindos/betterpvp/core/combat/damagelog/DamageLogListener.java
braulio-dev: Did it for cleanliness since I saw other classes like commands avoid creating a constructor if Guice can directly inject into their field member. I can remove it

core/src/main/java/me/mykindos/betterpvp/core/combat/events/CustomDamageReductionEvent.java
braulio-dev: Actually, no. The main event's damage may be calculated after by a custom damage adapter like mythic mobs or any we introduce later. The `initialDamage` in this method saves the latest real damage the player takes before any modifier like armor reduction or other adapters is introduced.

Mykindos: But isn't that the point of rawdamage in customdamageevent, to indicate what the initial damage was before any modifiers

braulio-dev: That's the raw damage of the attack. After that, comes modifiers like skills, then finally comes adapters and armor reduction. The modifiers is the `initialDamage` or the `finalDamage` of the custom damage events that is later overwritten by the adapters and armor reduction since there's no selective field for it.

braulio-dev: Capture (12ORm)

core/src/main/java/me/mykindos/betterpvp/core/combat/damagelog/DamageLogListener.java
Mykindos: Burning the steak, it's only commands that do it I believe

PR #378, opened 2024-01-21
Cannons, weapons, and fixes

braulio-dev:
Changes:
- Add [Cannons](https://youtu.be/AF8K8B786og):
    - Rotatable model. Hold right-click to aim
    - Loadable with cannonball. Shift + right-click to shoot
    - Custom cannon model.
    - Configurable fuse, shoot and cooldown times.
    - Configurable setting to explode cannonballs upon entity collision.
    - Configurable setting to explode cannonballs upon block collision.
    - Add cannon cannonballs to `ClansExplosionListener`
- Fix some skills not calling a velocity event and therefore not being able to be canceled.
- Fix player blocking state not working if the paper world configuration had changed the shield blocking delay.
- Make `SoundEffect` work with Kyori-Adventure types.
- Fix shops module removing _every_ entity upon server shutdown (?)
- Make `ConfigInjectorModule` not attempt to inject if the class is tagged with a `PluginAdapter` and it cannot be loaded.
- Make `ListenerLoader` not attempt to inject if the class is tagged with a `PluginAdapter` and it cannot be loaded.
- Change `projectile` to `damagingEntity` in `CustomDamageEvent` for listeners to know exactly what entity hit players. (TNT, in the case of cannons).
-  Move weapon system to `Core`.

Changes were tested.


champions/src/main/java/me/mykindos/betterpvp/champions/Champions.java
```diff
@@ -75,13 +77,15 @@ public void onEnable() {
...
-
             var recipeHandler = injector.getInstance(RecipeHandler.class);
             recipeHandler.loadConfig(this.getConfig(), "champions");
 
             updateEventExecutor.loadPlugin(this);
+
+            final Adapters adapters = new Adapters(this);
+            final Reflections reflectionAdapters = new Reflections(PACKAGE);
+            adapters.loadAdapters(reflectionAdapters.getTypesAnnotatedWith(PluginAdapter.class));
+            adapters.loadAdapters(reflectionAdapters.getTypesAnnotatedWith(PluginAdapters.class));
```
Mykindos, line 88: Is this still necessary

clans/src/main/java/me/mykindos/betterpvp/clans/weapons/ClansWeaponManager.java
```diff
@@ -0,0 +1,30 @@
...
+        this.adapters = new Adapters(clans);
+        this.clans = clans;
+    }
+
+    public void load() {
+        new Reflections(getClass().getPackageName()).getSubTypesOf(Weapon.class).forEach(clazz -> weaponManager.load(clans, adapters, clazz));
+    }
+
+}
+
```
Mykindos, line 30: same deal is championsweaponmanager

champions/src/main/java/me/mykindos/betterpvp/champions/weapons/ChampionsWeaponManager.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.google.inject.Inject;
+import me.mykindos.betterpvp.champions.Champions;
+import me.mykindos.betterpvp.core.combat.weapon.Weapon;
+import me.mykindos.betterpvp.core.combat.weapon.WeaponManager;
+import me.mykindos.betterpvp.core.framework.adapter.Adapters;
+import me.mykindos.betterpvp.core.framework.adapter.PluginAdapter;
+import org.reflections.Reflections;
+
+@PluginAdapter("Core")
```
Mykindos, line 11: Not sure why we need a PluginAdapter for core? seeing as all plugins are hard-dependent on it.

Other than that, this class just seems like a loader similar ChampionsCommandLoader & ChampionsListenerLoader

Mykindos:
+ has conflicts

champions/src/main/java/me/mykindos/betterpvp/champions/Champions.java
braulio-dev: yes, it loads plugin adapters without explicitly calling the class which loads its definition and could cause exceptions if we reference it directly

champions/src/main/java/me/mykindos/betterpvp/champions/weapons/ChampionsWeaponManager.java
braulio-dev: i only did that so it gets loaded without me having to call its constructor, how do you want me to load it

Mykindos (approved):

PR #459, opened 2024-01-28
Fix durability issues

### LINKED ISSUE #451: Dealing damage with abilities reduces durability(intended?)

StudentAlleg:
fixes #451, durability being consumed on all damage causes
Fixes bows taking double durability damage when causing damage


Mykindos:
Fixed this + some other stuff, credited as student as most of the work was already done here

PR #485, opened 2024-02-01
Damge falloff to cannonballs

### LINKED ISSUE #386: add falloff damage to tnt
- damage tnt deals should scale down from a max to a min depending on how far the player was from the tnt when it exploded

StudentAlleg:
Fixes #386


clans/src/main/java/me/mykindos/betterpvp/clans/weapons/impl/cannon/listener/CannonListener.java
```diff
@@ -145,7 +153,16 @@ public void onCustomDamage(final PreCustomDamageEvent pre) {
...
+                double distance = tnt.getLocation().distance(event.getDamagee().getLocation());
+                double damage;
+                if (distance < cannonballDamageMinRadius) {
+                    damage = cannonballDamage;
+                } else if (distance > cannonballDamageMaxRadius) {
+                    damage = 0;
+                } else {
+                    damage = cannonballDamage * (distance - cannonballDamageMinRadius) / cannonballDamageMaxRadius - cannonballDamageMinRadius;
+                }
+                event.setDamage(damage);
```
Mykindos, line 165: <= and >=, what if they are exactly 4 blocks away

If cannon damage is 15, and a player is 3.99 blocks away
15 * (3.99 - 1) / 4 - 1 = 10.2 damage, i feel like this could be simplified a bit and maybe reduce damage by more depending on distance

Mykindos (changes requested):

clans/src/main/java/me/mykindos/betterpvp/clans/weapons/impl/cannon/listener/CannonListener.java
StudentAlleg: Math was not mathing, math was fixed. <= and >= is not necessary, as the scaling function works on those numbers.

Mykindos: image

I would be ok with this, but we probably want to deal at least some damage if they were caught in the radius

StudentAlleg: What if it was damage = Math.max(cannonballDamage * ((deltaRadius - (distance - cannonballDamageMinRadius)) / deltaRadius), 1);

StudentAlleg: at max range, you'd at least deal 1 damage

Mykindos: lets make the minimum 4 (2 hearts to a naked)

StudentAlleg: implemented

PR #504, opened 2024-02-04
Add enable/disable to custom weapons

### LINKED ISSUE #407: Add a way to disable weapons/custom items

StudentAlleg:
Add enable config to weapons, prevent weapons from taking custom effects if not enabled.

Fixes #407

- [x] I have added spirits to weapons


core/src/main/java/me/mykindos/betterpvp/core/combat/weapon/Weapon.java
```diff
@@ -41,6 +56,64 @@ public void loadWeapon(BPVPItem item) {
...
+        return plugin.getConfig(getConfigName()).getObject(getPath(name), type, defaultValue);
+    }
+
+    protected String getConfigName() {
+        String weaponType = "standard";
+        if (this instanceof LegendaryWeapon) {
+            weaponType = "legendaries";
+        }
+        return "weapons/" + weaponType;
+    }
```
Mykindos, line 87: I'd prefer a default implementation in IWeapon which return "weapons/standard", then another default implementation in the LegendaryWeapon interface which would be weapons/legendaries. This will give us the ability to override it easily without making this function longer and longer

Mykindos (changes requested):

core/src/main/java/me/mykindos/betterpvp/core/combat/weapon/Weapon.java
StudentAlleg: Makes sense, implemented

Mykindos (approved):

PR #506, opened 2024-02-05
made brute passives the same layout as mineplex

ezrn:


Mykindos:
check if you need to update loadDefaultBuilds

ezrn:
done

Mykindos (approved):

PR #507, opened 2024-02-05
made knight passives same layout as mineplex

ezrn:


Mykindos:
check if you need to update loadDefaultBuilds in BuildRepository.java

ezrn:
done

PR #524, opened 2024-02-07
misc fixes

ezrn:
- Fixed blizzard not applying slowness for correct duration and having wrong spread
- Updated blizzard config to correct energy value
- Fixed seismic slam doing far more damage than intended(did 6 max on mineplex, 11 max on here)
- Gave inferno proper spread and increased its cooldown


Mykindos:
Defensive stance is just an internal cooldown, works as expected

ezrn:
> Defensive stance is just an internal cooldown, works as expected

Does not make any semblance of sense on a skill like that where you can simply hold it down for a set duration and then instantly use it again. If the internal cooldown is going to proc when the ability starts channeling it needs to be longer

Mykindos (approved):

PR #554, opened 2024-02-12
Void mage fix

### LINKED ISSUE #538: Void needs to reduce every based on damage not based on hits

ezrn:
- made void reduce energy based on damage dealt to the player instead of per hit
- Fixes #538 

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/mage/passives/Void.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/mage/passives/Void.java
@@ -51,7 +51,7 @@ public class Void extends ActiveToggleSkill implements EnergySkill {
                 "<effect>Slownesss " + UtilFormat.getRomanNumeral(slownessStrength + 1) + "</effect>, and take no knockback",
                 "",
                 "Every point of damage you take will be",
-                "reduced by <val" + getDamageReduction(level) + "</val> and drain <val>" + getEnergyReduction(level) + "</val> energy",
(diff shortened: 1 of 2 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/mage/passives/Void.java
```diff
@@ -48,10 +48,10 @@ public String[] getDescription(int level) {
                 "Drop your Sword / Axe to toggle",
                 "",
                 "While in void form, you receive",
-                "<effect>Slownesss " + UtilFormat.getRomanNumeral(slownessStrength + 1) + "</effect>, and take no Knockback",
+                "<effect>Slownesss " + UtilFormat.getRomanNumeral(slownessStrength + 1) + "</effect>, and take no knockback",
                 "",
-                "Reduces incoming damage by <stat>" + getDamageReduction(level) + "</stat>,",
-                "but burns <stat>" + getEnergyReduction(level) + "</stat> of your energy",
+                "Every point of damage you take will be",
+                "reduced by <val" + getDamageReduction(level) + "</val> and drain <val>" + getEnergyReduction(level) + "</val> energy",
```
Mykindos, line 54: evidently untested

Mykindos (changes requested):

Mykindos:
you also forgot to mention you reduced the amount of energy void consumes by 85%

ezrn:
> you also forgot to mention you reduced the amount of energy void consumes by 85%

this is to account for the fact that every point of damage reduces energyReduction energy, where before it was every hit. This means that if someone deals 6 damage it will be 6 x energyReduction which is at base 6 x 3.0 = 18, where it was 20 per hit before. It is practically the same.

Mykindos (approved):

PR #564, opened 2024-02-16
mage fixes

ezrn:
- Fixed static lazer not reducing energy per level
- Fixed inferno spread(forgot to add a 0 last time)
- Made inferno use energy, decreased cooldown back to original
- Fixed blizzard energy
- Fixed glacial blade not consuming energy

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/mage/sword/Inferno.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/mage/sword/Inferno.java
@@ -12,6 +12,7 @@ import me.mykindos.betterpvp.champions.champions.skills.data.ChargeData;
 import me.mykindos.betterpvp.champions.champions.skills.data.SkillActions;
 import me.mykindos.betterpvp.champions.champions.skills.types.ChannelSkill;
 import me.mykindos.betterpvp.champions.champions.skills.types.CooldownSkill;
+import me.mykindos.betterpvp.champions.champions.skills.types.EnergySkill;
 import me.mykindos.betterpvp.champions.champions.skills.types.InteractSkill;
 import me.mykindos.betterpvp.core.client.gamer.Gamer;
 import me.mykindos.betterpvp.core.combat.events.CustomDamageEvent;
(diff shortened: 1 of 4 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/mage/sword/StaticLazer.java
```diff
@@ -141,6 +140,7 @@ public void loadSkillConfig(){
         energyPerSecond = getConfig("energyPerSecond", 24.0, Double.class);
         collisionRadius = getConfig("collisionRadius", 1.8, Double.class);
         explosionRadius = getConfig("explosionRadius", 4.0, Double.class);
+        energyPerSecondReductionPerLevel = getConfig("energyPerSecondReductionPerLevel", 2.0, Double.class);
```
Mykindos, line 143: I think adding energyPerSecond and energyPerSecondReductionPerLevel were a mistake here (probably reys fault)

All energy skills have a default configuration

image

Can you change it to use these instead

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/mage/sword/StaticLazer.java
ezrn: energyPerSecondReductionPerLevel is for the energy consumed per second, the skill also has an initial energy cost which is defined by energy and energyDecreasePerLevel

Mykindos: No other skill does that, we typically just use the base energy cost as the activation cost, or pick a relatively low number

ezrn: well both static and now this do. I can change them both to use a higher amount and drain over time

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/mage/sword/Inferno.java
```diff
@@ -85,14 +86,19 @@ public String[] getDescription(int level) {
...
     public double getFireDuration(int level) {
         return baseFireDuration + level * fireDurationIncreasePerLevel;
     }
 
+    private float getEnergyPerSecond(int level) {
+        return (float) (energy - ((level - 1) * energyDecreasePerLevel));
+    }
+
     public int getNumFlames(int level){
         return baseNumFlames + level * numFlamesIncreasePerLevel;
```
Mykindos, line 103: fix the scaling here to be level -1, same with getFireDuration above (i know you didnt change those in this PR but might as well fix

Mykindos (changes requested):

Mykindos (approved):

Issue #426, opened 2024-01-24
revert roped arrow to mineplex values and give it a particle trail/effect that resembles a rope


ezrn:

ezrn:
I got this

braulio-dev:
Fixed with #565

PR #565, opened 2024-02-16
made roped arrow like mineplex

ezrn:
- Shorter cooldown
- Much less velocity
- Enabled its use while slowed since its much weaker now
- Fixes most of #426 


braulio-dev:
Attaching a leash to an arrow requires a tad bit more work so we'll go ahead and put that off for later

PR #567, opened 2024-02-16
allowed wolfs pounce to charge while jumping

ezrn:
- Still is unusable in the air, but when jumping on flat ground or while hanging off sides of blocks it will correctly charge


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/sword/WolfsPounce.java
```diff
@@ -283,8 +283,11 @@ public void updateCharge() {
 
             // Check if they still are blocking and charge
             if (isHolding(player) && gamer.isHoldingRightClick()) {
-                // Reset their charge if they're not on the ground
-                if (!UtilBlock.isGrounded(player)) {
+                // Check if the player is grounded or the block directly beneath them is solid
+                boolean isDirectlyAboveGround = !UtilBlock.airFoliage(player.getLocation().add(0, -1, 0).getBlock());
+                boolean isOneBlockAboveGround = !UtilBlock.airFoliage(player.getLocation().add(0, -2, 0).getBlock());
```
Mykindos, line 288: UtilBlock.isGrounded takes an additional argument which is the amount of blocks to check for, so I think UtilBlock.isGrounded(player, 2) will work for you here

Mykindos (changes requested):

PR #613, opened 2024-02-19
Stampede fix

ezrn:
- Fixed lots of stuff with stampede
- Unspaghettified the code


Mykindos:
you somehow made it more spaghetti despite ur claim or unspagettifying the code

ezrn:
> you somehow made it more spaghetti despite ur claim of unspagettifying the code

how is it more spaghetti

Mykindos:
> > you somehow made it more spaghetti despite ur claim of unspagettifying the code
> 
> how is it more spaghetti? It is much more readable than it used to be

you added 2 more hashmaps lol

ezrn:
> > > you somehow made it more spaghetti despite ur claim of unspagettifying the code
> > 
> > 
> > how is it more spaghetti? It is much more readable than it used to be
> 
> you added 2 more hashmaps lol

sprintingPlayers is necessary to fix a bug that has apparently always existed with stampede where canceling your sprint doesn't reset your sprint time on the first level, and active players is necessary to reduce server load by only iterating through players with the active skill instead of all online players like it used to. Only channel skills have an inherited active map so I had to make my own

Mykindos:
> > > > you somehow made it more spaghetti despite ur claim of unspagettifying the code
> > > 
> > > 
> > > how is it more spaghetti? It is much more readable than it used to be
> > 
> > 
> > you added 2 more hashmaps lol
> 
> sprintingPlayers is necessary to fix a bug that has apparently always existed with stampede where canceling your sprint doesn't reset your sprint time on the first level, and active players is necessary to reduce server load by only iterating through players with the active skill instead of all online players like it used to. Only channel skills have an inherited active map so I had to make my own

Looping through all players isn't bad if you are skipping the entries that aren't applicable quickly @ low cost, e.g. isSprinting or getLevel are quick checks.

I'm not sure what the bugs you are talking about are, but you dont need to do what youve done to fix them

ezrn:
> > > > > you somehow made it more spaghetti despite ur claim of unspagettifying the code
> > > > 
> > > > 
> > > > how is it more spaghetti? It is much more readable than it used to be
> > > 
> > > 
> > > you added 2 more hashmaps lol
> > 
> > 
> > sprintingPlayers is necessary to fix a bug that has apparently always existed with stampede where canceling your sprint doesn't reset your sprint time on the first level, and active players is necessary to reduce server load by only iterating through players with the active skill instead of all online players like it used to. Only channel skills have an inherited active map so I had to make my own
> 
> Looping through all players isn't bad if you are skipping the entries that aren't applicable quickly @ low cost, e.g. isSprinting or getLevel are quick checks.
> 
> I'm not sure what the bugs you are talking about are, but you dont need to do what youve done to fix them

rey told me to always use a map to loop through instead of players. The bug im talking about is that stampede needs to remove playres from the other maps the second they stop sprinting. In the previous code it was checking if the player was sprinting, and the level was > 0, which doesnt work since for the first ramp up to speed the playres level was 0, meaning that as long as the player had sprinted at some time before, it would instantly put them at speed 1 the second they started sprinting. The map fixes this

Mykindos:
> rey told me to always use a map to loop through instead of all online players. The bug im talking about is that stampede needs to remove playres from the other maps the second they stop sprinting. In the previous code it was checking if the player was sprinting, and the str was > 0, which doesnt work since for the first ramp up to speed the playres level was 0, meaning that as long as the player had sprinted at some time before, it would instantly put them at speed 1 the second they started sprinting. The map fixes this

https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/event/player/PlayerToggleSprintEvent.html 🙈

Mykindos:
This skill should be do-able with a single HashMap, storing an object called StampedeData or something

ezrn:
> This skill should be do-able with a single HashMap, storing an object called StampedeData or something

I can consolidate everything into 1 map, though I'm still not sure what the issue is with having multiple maps for a class. Does it cause performance issues?

Mykindos:
> > This skill should be do-able with a single HashMap, storing an object called StampedeData or something
> 
> I can consolidate everything into 1 map, though I'm still not sure what the issue is with having multiple maps for a class. Does it cause performance issues?

It creates unnecessary overhead, makes the code harder to maintain and less readable

The simpler something can be, the better

ezrn:
fixed

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/passives/Stampede.java
```diff
@@ -166,19 +149,22 @@ public void removeSpeed(Player player) {
...
         Player player = event.getPlayer();
-        if (!activePlayers.containsKey(player)) return;
-
         boolean isSprinting = event.isSprinting();
-        StampedeData data = playerData.getOrDefault(player, new StampedeData(System.currentTimeMillis(), 0));
+
+        StampedeData data = playerData.get(player);
+        if (data == null) {
+            data = new StampedeData(System.currentTimeMillis(), 0);
+        }
```
Mykindos, line 157: getOrDefault accomplished the same thing for you here

Mykindos (changes requested):

PR #616, opened 2024-02-20
Ranger sharpshooter and other misc changes

ezrn:
- Made sharpshooter how it was on mineplex(lower damage per shot, higher cap, time expiration instead of hit expiration)
- Added sounds and feedback messages to sharpshooter
- Added custom ranger arrow hit sound similar to mineplex
- Added particles to agility
- Fixed particles on hold position


champions/src/main/java/me/mykindos/betterpvp/champions/listeners/ArrowListener.java
```diff
@@ -47,6 +56,18 @@ public void onShootBow(EntityShootBowEvent event) {
...
+        if (!(event.getEntity() instanceof Arrow arrow)) return;
+        if (!(arrow.getShooter() instanceof Player player)) return;
+        if (!(roleManager.hasRole(player, Role.RANGER))) return;
+        Entity hitEntity = event.getHitEntity();
+        if (hitEntity instanceof LivingEntity damagee) {
+            if(damagee instanceof Player && ((Player) damagee).getGameMode() == GameMode.CREATIVE) return;
+            damagee.getWorld().playSound(damagee.getLocation(), Sound.ENTITY_PLAYER_HURT, 1f, 0.5f);
+        }
+    }
+
```
Mykindos, line 70: https://github.com/Mykindos/BetterPvP/blob/77bfd2bd57fd6b92da030daf2d69f1501e2cb4c8/core/src/main/java/me/mykindos/betterpvp/core/combat/listeners/CombatListener.java#L173-L174

ezrn: Intended, some arrows on ranger like sharpshooter, longshot, and overcharge had this sound and it made the arrows sounds much more powerful when they hit, as fits with ranger compared to assassin which is supposed to feel weaker.

Mykindos: this isn't overcharge though, this is every arrow

ezrn: yes this is every arrow shot by a ranger, as ranger arrows are more potent and powerful than assassin arrows. If you would like I could only put it onto overcharge, longshot, sharpshooter, and precision, but I think its nice to have an extra indicator that tells the player they are being shot by a ranger arrow vs another type of arrow

Mykindos: would prefer if you just removed it, really dont need more sounds

ezrn: can I add it as the death sound. it was also the death sound on mineplex

PR #631, opened 2024-02-27
Vulnerability Changes

ezrn:
- Each level now increments by 10% starting at 10%
- This allows variable levels of vulnerability to be used without being stuck with +50% increase or +25% increase, one of which is weak, and the other which is way too strong

### DIFF
--- a/core/src/main/java/me/mykindos/betterpvp/core/effects/listeners/EffectListener.java
+++ b/core/src/main/java/me/mykindos/betterpvp/core/effects/listeners/EffectListener.java
@@ -227,7 +227,7 @@ public class EffectListener implements Listener {
         Optional<Effect> effectOptional = effectManager.getEffect(event.getDamagee(), EffectType.VULNERABILITY);
         effectOptional.ifPresent(effect -> {
             if (event.getCause() == EntityDamageEvent.DamageCause.LIGHTNING) return;
-            double damageIncrease = 0.25 + (effect.getLevel() - 1) * 0.10;
+            double damageIncrease = 0.1 + (effect.getLevel() - 1) * 0.10;
             event.setDamage(event.getDamage() * (1.0 + damageIncrease));
         });
     }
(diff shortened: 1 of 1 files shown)


Mykindos:
Feel like you could just make it like strength, 1.5 damage per level

StudentAlleg:
> Feel like you could just make it like strength, 1.5 damage per level

That would be incredibly strong with low damage sources, as vulnerability applies to all damage.

ezrn:
> > Feel like you could just make it like strength, 1.5 damage per level
> 
> That would be incredibly strong with low damage sources, as vulnerability applies to all damage.

Inferno vulnerability would 1 shot brutes with a full burst

ezrn:
There may be a better way to do this, though I think percentages are the way to go. Maybe starting vulnerability at a lower value like 10%, and having it go up by 10% each time, then compensating any current abilities that use Vulnerability with these new levels. MArked for death for instance is pretty weak atm, whereas during season 1 of the 2020 wipe is was crazy broken. A system like this to give in between values would be very valuable.

Mykindos:
> There may be a better way to do this, though I think percentages are the way to go. Maybe starting vulnerability at a lower value like 10%, and having it go up by 10% each time, then compensating any current abilities that use Vulnerability with these new levels. Marked for death for instance is pretty weak atm, whereas during season 1 of the 2020 wipe it was crazy broken. A system like this to give in between values would be very valuable.

I guess since sword damage was decreased, marked for death lost value
I'm not opposed to making it 10% and moving marked for death up to Vuln 3 or 4

core/src/main/java/me/mykindos/betterpvp/core/effects/listeners/EffectListener.java
```diff
@@ -227,7 +227,8 @@ public void onDamage(CustomDamageEvent event) {
         Optional<Effect> effectOptional = effectManager.getEffect(event.getDamagee(), EffectType.VULNERABILITY);
         effectOptional.ifPresent(effect -> {
             if (event.getCause() == EntityDamageEvent.DamageCause.LIGHTNING) return;
-            event.setDamage((event.getDamage() * (1.0 + (effect.getLevel() * 0.25))));
+            double damageIncrease = 0.1 + (effect.getLevel() - 1) * 0.10;
```
Mykindos, line 230: because this isn't a vanilla effect, you shouldnt do -1

You could keep it the same as it was before but replace 0.25 with 0.10

Mykindos (changes requested):

braulio-dev:
Deferred until @Mykindos pushes effect system rework

PR #638, opened 2024-02-29
Toggle skill drop QOL

### LINKED ISSUE #627: Allow Legends to activate passive abilities by dropping

### LINKED ISSUE #632: Allow users to drop swords/axes from inventory from inventory view
InventoryClickEvent has an ActionType that has DROP. This also triggers the normal drop event. It should be possible to allow the drop event if it comes from the inventory.

This would make it easier to do inventory management when using a class that has a drop skill, while still retaining that functionality.

StudentAlleg:
Allow players to drop items from inventory view without triggering toggle skills
Allow legends to toggle toggle skills

Fixes #632 
Fixes #627 

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/listeners/SkillListener.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/listeners/SkillListener.java
@@ -83,7 +83,7 @@ public class SkillListener implements Listener {
     private final SkillManager skillManager;
     private final WeaponManager weaponManager;
 
-    private final HashSet<Player> InventoryDrop = new HashSet<>();
+    private final HashSet<UUID> InventoryDrop = new HashSet<>();
 
     @Inject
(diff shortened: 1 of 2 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/listeners/SkillListener.java
```diff
@@ -76,17 +81,21 @@ public class SkillListener implements Listener {
     private final EffectManager effectManager;
     private final ClientManager clientManager;
     private final SkillManager skillManager;
+    private final WeaponManager weaponManager;
+
+    private final HashSet<Player> InventoryDrop = new HashSet<>();
```
Mykindos, line 86: store a unique ID instead,

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/listeners/SkillListener.java
StudentAlleg: fixed

PR #639, opened 2024-02-29
Add drop protection for UUIDitems

### LINKED ISSUE #609: Add drop protection for important items.
Add a setting that forces a confirmation when dropping items such as legendaries.

StudentAlleg:
Prevent dropping UUIDItems unless you drop it twice within a specified time period (currently 500ms).

Fixes #609

### DIFF
--- a/core/src/main/java/me/mykindos/betterpvp/core/client/properties/ClientProperty.java
+++ b/core/src/main/java/me/mykindos/betterpvp/core/client/properties/ClientProperty.java
@@ -9,6 +9,7 @@ public enum ClientProperty {
 
     SIDEBAR_ENABLED,
     TIPS_ENABLED,
+    DROP_PROTECTION_ENABLED,
     STAFF_CHAT,
     COOLDOWN_DISPLAY,
     CHAT_ENABLED,
--- a/core/src/main/java/me/mykindos/betterpvp/core/client/repository/ClientListener.java
+++ b/core/src/main/java/me/mykindos/betterpvp/core/client/repository/ClientListener.java
@@ -197,7 +197,10 @@ public class ClientListener implements Listener {
         if(tipsOptional.isEmpty()){
             client.saveProperty(ClientProperty.TIPS_ENABLED, true);
         }
-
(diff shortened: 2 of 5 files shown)


core/src/main/java/me/mykindos/betterpvp/core/items/listener/UuidListener.java
```diff
@@ -91,13 +95,33 @@ public class UuidListener implements Listener {
...
         this.clientManager = clientManager;
     }
 
+    @EventHandler(priority = EventPriority.HIGHEST)
+    public void onUUIDItemDrop(PlayerDropItemEvent event) {
+        if (event.isCancelled()) return;
+        Client client = clientManager.search().online(event.getPlayer());
+        if ((boolean) client.getProperty(ClientProperty.DROP_PROTECTION_ENABLED).orElse(false)) {
+            itemHandler.getUUIDItem(event.getItemDrop().getItemStack()).ifPresent(uuidItem -> {
+                if (!(lastUUIDDropTime.containsKey(uuidItem.getUuid()) && !UtilTime.elapsed(lastUUIDDropTime.get(uuidItem.getUuid()), UUIDdropTimeMs))) {
```
Mykindos, line 113: if map doesnt contain uuid and time has elapsed?

Could probably clean this up a bit, not very readable

Mykindos (dismissed):

core/src/main/java/me/mykindos/betterpvp/core/items/listener/UuidListener.java
StudentAlleg: If the map doesn't contain the UUID, then this is the first time (this session) that we have seen this legend, therefore it is impossible for time to have elasped.

I was trying to avoid unnecessary/nested if statements, but I can see it being confusing

StudentAlleg: Cleaned up

Mykindos: much better

braulio-dev (approved):

PR #657, opened 2024-03-08
made wolfs fury expire when switching off weapon

ezrn:


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/axe/WolfsFury.java
```diff
@@ -114,6 +114,10 @@ private boolean expire(Player player, boolean force) {
         if (player == null) {
             return true;
         }
+        if (!isHolding(player)) {
+            expire(player, true);
```
Mykindos, line 118: this will create a StackOverflow exception. call deactivate instead

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/axe/WolfsFury.java
```diff
@@ -114,6 +114,10 @@ private boolean expire(Player player, boolean force) {
...
+            deactivate(player);
+        }
```
Mykindos, line 120: i mean just have it as

`
 if (!isHolding(player)) {
          deactivate(player);
          return true;
 }
`

Mykindos (changes requested):

PR #664, opened 2024-03-09
Thorns rework + knight fixes

ezrn:
- Thorns now returns damage every 3 attacks within a timeframe(Default 2.0) Instead of returning damage every time you are hit with an internal cooldown(was 2.0)
- This makes the skill viable for breaking out of combos, but gives it counterplay because the enemies can wait to perform their third attack and completely negate the skill. 
- Also feels much better for the player since they dont have to wait for an arbitrary internal cooldown


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/passives/Thorns.java
```diff
@@ -67,31 +73,37 @@ public SkillType getType() {
...
+
+        LivingEntity damager = event.getDamager();
+        hitTimestamps.putIfAbsent(player, new LinkedList<>());
+        Queue<Long> timestamps = hitTimestamps.get(player);
+
+        timestamps.offer(System.currentTimeMillis());
+
+        while (!timestamps.isEmpty() && System.currentTimeMillis() - timestamps.peek() > (reflectTime * 1000)) {
+            timestamps.poll();
+        }
```
Mykindos, line 93: I would rather you just do this in an update event and process 1 per player per tick or 2 ticks.

Since LivingEntity is also nullable, migrate to a weakhashmap

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/passives/Thorns.java
ezrn: Just curious, whats problematic about this approach?

Mykindos: While you have done it in a safe way, while loops can cause the server to hang if they get stuck iterating. Since minecraft runs mainly on a single thread, I think its a safer bet to avoid them all together

Mykindos: This doesn't just apply to while loops, but any kind of loop

ezrn: > This doesn't just apply to while loops, but any kind of loop. Typically we loop over a fixed size or an existing list, but in this case you would be relying on the timestamp to be correctly removed before it exits

in that case there are definitely a few while loops in other classes that should be updated in this way.

PR #669, opened 2024-03-09
Rework effect system, add resistance global passive

Mykindos:
## Describe your changes
Heavily reworked the effect system
Added Resistance global passive

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/global/Resistance.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/global/Resistance.java
@@ -1,0 +1,100 @@
+package me.mykindos.betterpvp.champions.champions.skills.skills.global;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import me.mykindos.betterpvp.champions.Champions;
+import me.mykindos.betterpvp.champions.champions.ChampionsManager;
+import me.mykindos.betterpvp.champions.champions.skills.Skill;
+import me.mykindos.betterpvp.champions.champions.skills.types.PassiveSkill;
+import me.mykindos.betterpvp.core.components.champions.Role;
+import me.mykindos.betterpvp.core.components.champions.SkillType;
+import me.mykindos.betterpvp.core.effects.events.EffectReceiveEvent;
+import me.mykindos.betterpvp.core.listener.BPvPListener;
+import me.mykindos.betterpvp.core.utilities.UtilEffect;
+import me.mykindos.betterpvp.core.utilities.UtilServer;
+import org.bukkit.entity.Player;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.EventPriority;
+import org.bukkit.event.entity.EntityPotionEffectEvent;
+import org.bukkit.potion.PotionEffect;
+
+@Singleton
+@BPvPListener
+public class Resistance extends Skill implements PassiveSkill {
+
+    private double baseDurationReduction;
+
+    private double durationReductionPerLevel;
+
+    @Inject
+    public Resistance(Champions champions, ChampionsManager championsManager) {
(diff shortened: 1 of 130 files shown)


braulio-dev (approved):

core/src/main/java/me/mykindos/betterpvp/core/effects/listeners/effects/ResistanceListener.java
```diff
@@ -0,0 +1,35 @@
...
+    @Inject
+    public ResistanceListener(EffectManager effectManager) {
+        this.effectManager = effectManager;
+    }
+
+    @EventHandler(priority = EventPriority.HIGHEST)
+    public void resistanceReduction(CustomDamageEvent event) {
+        if (event.getDamagee() instanceof Player player) {
+            Optional<Effect> effectOptional = effectManager.getEffect(player, EffectTypes.RESISTANCE);
+            effectOptional.ifPresent(effect -> event.setDamage(event.getDamage() * (1.0 - (effect.getAmplifier() * 20) * 0.01)));
```
StudentAlleg, line 32: The `20` represents base player health. This doesn't take into account say defensive aura, which I think is intentional (and wanted?).

StudentAlleg (commented):
The biggest note I have from code review, is that a lot of values (i.e. strength damage, resistance reduction) would make sense to have that defined in a config, instead of raw values.

core/src/main/java/me/mykindos/betterpvp/core/effects/listeners/effects/ResistanceListener.java
braulio-dev: This isn't 20 because of player base health, it's 20% per amplifier level (20 * 0.01 = 0.2 = 20%)

StudentAlleg:
For vanilla effects
Giving a player an infinite custom effect and it cannot be cleared. Give player a single custom effect and it cannot be cleared.
Give a player an infinite custom effect and a duration customeffect and it can be cleared only through the vanilla clear.
Tested with poison and regeneration.

Silence was able to be removed.

This behavior is somewhat inconsistent.

core/src/main/java/me/mykindos/betterpvp/core/effects/listeners/effects/ProtectionListener.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Inject
+    public ProtectionListener(EffectManager effectManager) {
+        this.effectManager = effectManager;
+    }
+
+    @EventHandler
+    public void entityDamageEvent(EntityDamageEvent event) {
+        if (event.getEntity() instanceof Player player) {
+            if (effectManager.hasEffect(player, EffectTypes.PROTECTION)) {
```
StudentAlleg, line 30: This stops all damage, including fall damage and mob damage. And presumably includes bosses (though probably will block them)

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/global/Resistance.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    public String getName() {
+        return "Resistance";
+    }
+
+    @Override
+    public String[] getDescription(int level) {
+
+        return new String[]{
+                "Negative effects have their duration reduced by <stat>" + getDurationReduction(level) + "%</stat>",
```
StudentAlleg, line 43: Should be <val> instead of <stat>

core/src/main/java/me/mykindos/betterpvp/core/effects/listeners/effects/ProtectionListener.java
Mykindos: ill probably tweak this when i properly implement the protection system (right now its never applied / command to remove it doesnt exist)

PR #674, opened 2024-03-11
Proposed Fortify Replacement

RockyMichael:
## Describe your changes
-Added Aerobatics: Increases damage when in the air. Works with both standard melee attacks and damage from abilities. This passive ability is better suited for brute, as it enables more offensive play, as opposed to fortify, which prolonged fights and doesn't quite fit with the identity of the kit.
N/A

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/passives/Aerobatics.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/passives/Aerobatics.java
@@ -1,0 +1,75 @@
+package me.mykindos.betterpvp.champions.champions.skills.skills.brute.passives;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import me.mykindos.betterpvp.champions.Champions;
+import me.mykindos.betterpvp.champions.champions.ChampionsManager;
+import me.mykindos.betterpvp.champions.champions.skills.Skill;
+import me.mykindos.betterpvp.champions.champions.skills.types.PassiveSkill;
+import me.mykindos.betterpvp.core.combat.events.CustomDamageEvent;
+import me.mykindos.betterpvp.core.components.champions.Role;
+import me.mykindos.betterpvp.core.components.champions.SkillType;
+import me.mykindos.betterpvp.core.listener.BPvPListener;
+import me.mykindos.betterpvp.core.utilities.UtilBlock;
+import org.bukkit.entity.Player;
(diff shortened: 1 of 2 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/passives/Aerobatics.java
```diff
@@ -0,0 +1,75 @@
...
+    public Role getClassType() {
+        return Role.BRUTE;
+    }
+
+    @EventHandler(priority = EventPriority.HIGHEST)
+    public void onDamage(CustomDamageEvent event) {
+        if (event.getDamager() instanceof Player damager) {
+            int level = getLevel(damager);
+            if (level > 0) {
+                if(UtilBlock.airFoliage(damager.getLocation().add(0,-1, 0).getBlock())){
```
StudentAlleg, line 57: This also seems to apply while jumping.

StudentAlleg, line 57: Distance above ground should be in config

StudentAlleg (changes requested):
There should be some way to tell that you are increasing damage with this skill.

Consider adding a cooldown.

I am not sure that this skill will work well, as in order to make is usable with seismic slam and takedown (the only two skills it makes sense with), it makes jumping and hitting very strong.

(not commenting on balance as suggested)

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/passives/Aerobatics.java
ezrn: should probobly be a isGrounded check instead, as the idea is that while you are airborne you deal more damage

RockyMichael: This is intended, as the purpose of the ability is to jump to deal extra damage. The trade-off is that it is easier to be comboed while you jump, so you have to think tactically about when you want to jump for the extra damage.

RockyMichael:
I agree that there should be a sound effect played to differentiate when the ability is used. Will add that.

ezrn:
getConfig calls use the wrong names

RockyMichael:
-Uses grounded check instead of checking 1 block below player for air. 
-Added sound to effect.
-Fixed config naming issues.

PR #723, opened 2024-05-04
Add Frozen effect and Punishment

### LINKED ISSUE #722: Add frozen effect
For staff/other tools.

- Prevents effectee from moving
- Prevents effectee from dropping
- Prevents effectee picking up items 
- Prevents effectee placing items
- Prevents effectee using items
- Prevents effectee using skills
- Prevents effectee from dealing damage
- Prevents effectee from taking damage

StudentAlleg:
Adds the frozen effect. While frozen, players cannot moving, attack, take damage, drop items, pickup items, break blocks, use skills, or anything else. They can chat.

The frozen punishment allows for staff to quickly prevent rule breaking players from continuing to act, before applying the appropriate punishment (a lot easier to freeze someone bunny hopping on someone, then type out the correct reason/time, rather than skip the frozen step).

Also allows staff to freeze a player and talk with them, for whatever reason.

- [x] Add frozen effect
- [x] Add frozen punishment

Fixes #722

### DIFF
--- a/core/src/main/java/me/mykindos/betterpvp/core/effects/listeners/effects/FrozenListener.java
+++ b/core/src/main/java/me/mykindos/betterpvp/core/effects/listeners/effects/FrozenListener.java
@@ -1,0 +1,97 @@
+package me.mykindos.betterpvp.core.effects.listeners.effects;
+
+import me.mykindos.betterpvp.core.effects.EffectManager;
+import me.mykindos.betterpvp.core.effects.EffectTypes;
+import me.mykindos.betterpvp.core.listener.BPvPListener;
+import me.mykindos.betterpvp.core.utilities.UtilMessage;
+import org.bukkit.entity.LivingEntity;
+import org.bukkit.entity.Player;
+import org.bukkit.event.Event;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.EventPriority;
+import org.bukkit.event.Listener;
+import org.bukkit.event.entity.EntityDamageByEntityEvent;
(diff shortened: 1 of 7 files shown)


core/src/main/java/me/mykindos/betterpvp/core/effects/listeners/effects/FrozenListener.java
```diff
@@ -0,0 +1,97 @@
...
+    @EventHandler
+    public void onPickupItem(PlayerAttemptPickupItemEvent event) {
+        if (event.isCancelled()) return;
+        if (effectManager.hasEffect(event.getPlayer(), EffectTypes.FROZEN)) {
+            event.setCancelled(true);
+        }
+    }
+
+    @EventHandler(priority = EventPriority.LOW)
+    public void onInteractEvent(PlayerInteractEvent event) {
```
Mykindos, line 60: Might want to check if this is already cancelled, and the hand the player interacted with. Might end up sending multiple messages otherwise

Mykindos (changes requested):

core/src/main/java/me/mykindos/betterpvp/core/effects/listeners/effects/FrozenListener.java
StudentAlleg: PlayerInteractEvent#isCancelled is deprecated, but I did change it to return if both of PlayerInteractEvent#useInteractedBlock and PlayerInteractEvent#useItemInHand are Event.Result.DENY (the full thing is cancelled)

Mykindos (approved):

PR #740, opened 2024-05-27
Bulls Eye Changes

ImMuscleMan:
## Describe your changes
 - Changed bulls eye particle spawn logic so that they only spawn when the player is charging or holding a bow (not in the way during combat)
 - Buffed Bulls Eye's curve distance by adding an extra 0.5 blocks of range which ignores current charge and making Bulls Eye 
 focus enemies faster


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/sword/Bullseye.java
```diff
@@ -138,9 +145,14 @@ public void updateCharge() {
                 continue;
             }
             // Spawn particles
+            if (player == null) return;
             if (playerBullsEyeData.getTargetFocused() != null && playerBullsEyeData.getTarget().isValid()) {
-                playerBullsEyeData.spawnFocusingParticles();
+                Gamer gamer = championsManager.getClientManager().search().online(player).getGamer();
```
Mykindos, line 150: You are already getting the gamer below this, no need to do it twice in the same function. Do it once just after the level check

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/sword/Bullseye.java
```diff
@@ -138,11 +145,15 @@ public void updateCharge() {
                 continue;
             }
             // Spawn particles
+            if (player == null) return;
```
Mykindos, line 148: Unnecessary check, its already being done here

`Bukkit.getPlayer(playerUUID) == null`

There is a little double up there but not too concerned about it

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/sword/Bullseye.java
```diff
@@ -216,17 +227,35 @@ public void onDamage(CustomDamageEvent event) {
...
         Player caster = playerBullsEyeData.getCaster();
         RayTraceResult result = caster.rayTraceEntities(64);
-        if (result == null || result.getHitEntity() == null) return;
-        if (!(result.getHitEntity() instanceof LivingEntity)) return;
-
-        if (!playerBullsEyeData.hasTarget()) {
-            playerBullsEyeData.setTarget((LivingEntity) result.getHitEntity());
-            playerBullsEyeData.setTargetFocused(new ChargeData((float)(0.35)));
-            bullsEyeData.put(caster.getUniqueId(), playerBullsEyeData);
+        if (result != null && result.getHitEntity() != null) {
```
Mykindos, line 230: Instead of checking result.getHitEntity() != null, you can just do this instead

```
if(result != null && !(result.getHitEntity() instanceof LivingEntity target)) {
           if (!playerBullsEyeData.hasTarget()) {
                playerBullsEyeData.setTarget(target);
                playerBullsEyeData.setTargetFocused(new ChargeData((float) (0.5)));
                bullsEyeData.put(caster.getUniqueId(), playerBullsEyeData);
                playerBullsEyeData.getTargetFocused().tick();
                playerBullsEyeData.getTargetFocused().tickSound(caster);
                playerBullsEyeData.updateColor();
            }
}
```

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/sword/Bullseye.java
ImMuscleMan: The code you suggested doesnt work
image

Mykindos: Try compile it anyway, theres no reason that shouldnt work. You can try set your IDE language level to java 21

StudentAlleg: isn't the `(result.getHitEntity() instanceof LivingEntity target)` being negated, so if getHitEntity is an instance of LivingEntity, the code will ever run (therefore there will never be a `target`). It doesn't look like it should be being negated at all.

Mykindos: We do this all over the place

image

Only reason I can think of it not working is his IDE thinking that the variable wasnt created

However it does raise a different question, this code is checking that the target is not a living entity, when really you want to check if it IS a living entity. Code should still compile either way though

StudentAlleg: but it's not returning before, this code relies on that boolean statement being true, if it is true, as written, target doesn't exist.

I think you introduced a logic error, if you remove the ! it should work.

Mykindos: Ye you right

Mykindos (approved):

PR #753, opened 2024-05-29
Update Mage Active Toggle Passives

### LINKED ISSUE #751: Energy drain for mage passives are inconesistent
https://www.youtube.com/watch?v=xYDQSWWTywM
Energy drain for mage passives are inconsistent with each other. 

Life Bonds:
Stated energy drain, 21.0/second
Takes ~10 seconds to fully drain

Inferno:
Stated energy drain, 7.0/second
Takes ~7 seconds to fully drain

Arctic Armor:
Stated energy drain, 6.0/second
Takes ~12 seconds to fully drain

Void:
Stated energy drain, 36.0/second
Takes ~4 seconds to fully drain

The difference between inferno and life bonds compared to their numbers is incorrect, not sure where the problem is currently.

StudentAlleg:
All Mage active toggle passives now correctly use config values in terms of seconds. Energy draining is now consistent across all skills. Config values are updated to keep the same effect they had previously. Add energyStartCost config, which is the cost for activating the skill.

Arctic armour now updates at the same rate as other active toggle passives.
Arctic armour now has a start cost.

Fixes #751

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/mage/passives/ArcticArmour.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/mage/passives/ArcticArmour.java
@@ -202,11 +202,6 @@ public class ArcticArmour extends ActiveToggleSkill implements EnergySkill {
     }
 
     @Override
-    public float getEnergyStartCost(int level) {
-        return (float) (energyStartCost - ((level - 1) * energyStartCostDecreasePerLevel));
-    }
-
-    @Override
     public void toggleActive(Player player) {
         if (championsManager.getEnergy().use(player, getName(), getEnergyStartCost(getLevel(player)), false)) {
             UtilMessage.message(player, getClassType().getName(), "Arctic Armour: <green>On");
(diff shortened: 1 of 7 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/mage/passives/ArcticArmour.java
```diff
@@ -199,9 +201,21 @@ public float getEnergy(int level) {
         return (float) (energy - ((level - 1) * energyDecreasePerLevel));
     }
 
+    @Override
+    public float getEnergyStartCost(int level) {
+        return (float) (energyStartCost - ((level - 1) * energyStartCostDecreasePerLevel));
+    }
```
Mykindos, line 207: Seeing as this is the same across all, I propose just overriding it in ActiveToggleSkill

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/mage/passives/ArcticArmour.java
StudentAlleg: I completely missed that ActiveToggleSkill extends skill (and therefore has access to the variables)

PR #771, opened 2024-05-30
#337 - Fixed Healing Shot

### LINKED ISSUE #337: healing shot does not always proc on self if charge is too low
this might not be fixable (or be important to fix), as the detection on self does things differently

manu10000:
## Describe your changes
Added a way for healing shot to proc when charge is low (e.g. when arrow doesnt leave the hitbox initially)

#337


manu10000:
probably need to remove arrows from the hashset in the onPreDamage event

Mykindos:
I still think an easier way to do this would be checking the velocity of the arrow

Mykindos (approved):

PR #785, opened 2024-05-31
#417 - Fix movement legends

### LINKED ISSUE #417: tooth doesn't work 100% of the time when you're bobbing up and down on water
[Showcase of when it doesn't work](https://youtu.be/fyrJvzfVz1w)

bearloo:
## Describe your changes
[Before](https://youtu.be/Rv4YYhUfX_c)
[After](https://youtu.be/g8aTTsNKnjo)

Updated movements to prevent canUse from removing players from the active set. Now it just returns before draining energy/applying abilities.
Since the player is still holding right click, the channel weapon check in [WeaponListener.java](https://github.com/Mykindos/BetterPvP/blob/master/core/src/main/java/me/mykindos/betterpvp/core/combat/weapon/listeners/WeaponListener.java) prevents them from becoming active again until they toggle right-click.

Relevant code from WeaponListener:

```
if (weapon instanceof ChannelWeapon channelWeapon) {
            if (clicked.get(event.getPlayer().getUniqueId()) == weapon) {
                return; // Skip if we're currently holding click on this weapon
            }

            if (channelWeapon.getEnergy() > 0) {
                if (!energyHandler.use(player, name, channelWeapon.getEnergy(), true)) {
                    return;
                }
            }

            clicked.put(player.getUniqueId(), weapon); // Log this weapon as clicked
}

if (weapon instanceof InteractWeapon interactWeapon) {
        interactWeapon.activate(player);
}
```

To-Do:
- [x] Reduce the number of logs produced by `canUse()`
- [x] Review gameplay change for Windblade, players can now fly out the of water if they enter swim mode (Same behavior seen in other skills so I think this is a non-issue)

Fixes #417

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/weapons/impl/legendaries/WindBlade.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/weapons/impl/legendaries/WindBlade.java
@@ -45,8 +45,6 @@ public class WindBlade extends ChannelWeapon implements InteractWeapon, Legendar
 
     private static final String ABILITY_NAME = "Flight";
 
-    private final Set<UUID> activeUsageNotifications = new HashSet<>();
-
     private double velocityStrength;
     private final EnergyHandler energyHandler;
     private final ClientManager clientManager;
@@ -75,20 +73,6 @@ public class WindBlade extends ChannelWeapon implements InteractWeapon, Legendar
     @Override
     public void activate(Player player) {
         active.add(player.getUniqueId());
-    }
-
-    @Override
-    @EventHandler
-    public void onDeath(PlayerDeathEvent event) {
-        activeUsageNotifications.remove(event.getPlayer().getUniqueId());
-        active.remove(event.getPlayer().getUniqueId());
-    }
-
-    @Override
-    @EventHandler
-    public void onQuit(PlayerQuitEvent event) {
-        activeUsageNotifications.remove(event.getPlayer().getUniqueId());
-        active.remove(event.getPlayer().getUniqueId());
     }
 
     @UpdateEvent (priority = 99)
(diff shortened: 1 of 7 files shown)


Mykindos:
Will take a look at this after alpha finishes

bearloo:
Reverted non-gameplay-specific changes. 

Changes included are for: 
- Ensuring legends remain active if `canUse()` fails, fixes some issues seen where legends fail to activate 
- Only sending one usage notification if `canUse()` fails multiple times in a row, resets if the legend becomes inactive or `canUse()` succeeds

Let me know your suggestions for refactoring the code duplication. 

- The notification logic belongs in InteractWeapon but I can't change the interface to an abstract class without breaking inheritance, is something like an "InteractableChannelWeapon" that implements a ChannelWeapon and InteractWeapon interface a viable solution?
- There is a lot of duplicated pre-condition code for checking a legend hasn't become inactive that could get moved up into a parent class shared between all movement legends (or any ChannelWeapon legendary really)

champions/src/main/java/me/mykindos/betterpvp/champions/weapons/impl/legendaries/WindBlade.java
```diff
@@ -30,15 +30,24 @@
...
+import java.util.HashSet;
+import java.util.Set;
 
 @Singleton
 @BPvPListener
 public class WindBlade extends ChannelWeapon implements InteractWeapon, LegendaryWeapon, Listener {
 
+    private static final String ABILITY_NAME = "Flight";
+
+    private final Set<UUID> activeUsageNotifications = new HashSet<>();
```
Mykindos, line 49: instead of creating this set for every ChannelWeapon, just created a protected set in ChannelWeapon itself

Mykindos (changes requested):

PR #820, opened 2024-06-03
Sever nerf

ezrn:
Makes sever deal 0.75 hearts per tick (down from 1)
Lasts 4 seconds (up from 3)
Total damage remains the same


manu10000:
If this ability is getting a nerd slash cooldown should also get a nerf

Mykindos:
Made changes myself as i didnt want to change the duration

PR #821, opened 2024-06-03
revert stampede speed

ezrn:
- Increases max speed to 3 (up from 2)
- Increases time between charges to 7 (up from 4)


Mykindos:
Not sure i want brute running faster than assassin after seeing it in action

PR #823, opened 2024-06-03
Add Basic Punish History Command

StudentAlleg:
Add basic punish history command. There isn't as much information as I'd like in it, but it's not worth adding that in during kit map. This will allow a temporary solution to allow admins to easily view the punish history of an account.


StudentAlleg:
https://prnt.sc/nLrkWBmKEt4N

Mykindos (approved):

PR #842, opened 2024-06-05
Fix certain passive abilities activating on killing Swarm bats from a Mage

ChakaClown:
Added checks within Null Blade, Bloodlust, Soul Harvest, and Cleave to not activate when attacking Bats, therefore removing the bugs where you can kill Swarm bats to activate these abilities.

I decided to not make the bats invincible so players could still theoretically use abilities to break through the swarm of bats to reach mage players.


Mykindos (changes requested):
I'm guessing you tried to do this through the web browser, as you didn't import Bat into the changed classes (and thus it doesn't compile)

But I wouldn't recommend doing it this way anyway, because what if we have another skill that spawns some mob type in the future?

What I propose doing is, in Swarm.java when the bat is spawned, call
`bat.setMetadata("PlayerSpawned", new FixedMetadataValue(champions, true));`

And then in all of these classes, instead of checking instanceof Bat, check entity.hasMetadata("PlayerSpawned")

Mykindos:
did this on my end as i wanted it patched tonight

PR #846, opened 2024-06-05
Cleave Nerf

ezrn:
- Now has a maximum of 3 blocks reach from the target enemy regardless of level


Mykindos (approved):
will need to tweak it more in the future to not go through blocks

PR #892, opened 2024-06-12
Fix Freeze Especially for Busjack (disable /kill while frozen)

StudentAlleg:
Prevent /kill while frozen

### DIFF
--- a/core/src/main/java/me/mykindos/betterpvp/core/command/commands/admin/FreezeCommand.java
+++ b/core/src/main/java/me/mykindos/betterpvp/core/command/commands/admin/FreezeCommand.java
@@ -43,7 +43,7 @@ public class FreezeCommand extends Command implements IConsoleCommand {
 
     @Override
     public void execute(Player player, Client client, String... args) {
-        if (player.getUniqueId().compareTo(UUID.fromString("90f3004d-ebb4-4b74-b30e-b320ad687256")) == 0) {
+        if (player.getUniqueId().equals(UUID.fromString("90f3004d-ebb4-4b74-b30e-b320ad687256"))) {
             effectManager.addEffect(player, EffectTypes.SHOCK, 10000);
             effectManager.addEffect(player, EffectTypes.DARKNESS, 10000);
             effectManager.addEffect(player, EffectTypes.BLINDNESS, 10000);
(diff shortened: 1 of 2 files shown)


core/src/main/java/me/mykindos/betterpvp/core/command/commands/admin/FreezeCommand.java
```diff
@@ -39,11 +43,17 @@ public String getDescription() {
 
     @Override
     public void execute(Player player, Client client, String... args) {
+        if (player.getUniqueId().compareTo(UUID.fromString("90f3004d-ebb4-4b74-b30e-b320ad687256")) == 0) {
```
Mykindos, line 46: someone has been doing c++ or something

Mykindos (changes requested):

core/src/main/java/me/mykindos/betterpvp/core/command/commands/admin/FreezeCommand.java
Mykindos: > I tried == and it didn't work :(

Because == is comparing the reference to the object, which is false because they are different objects
.equals on the other hand uses the objects equals comparator

There is also another options, convert the unique ID to string and then just compare strings

Mykindos: But despite how funny it could be, best to leave these things out :)

StudentAlleg: :(

core/src/main/java/me/mykindos/betterpvp/core/command/commands/admin/FreezeCommand.java
```diff
@@ -15,6 +16,7 @@
 import org.bukkit.entity.Player;
 
 @Singleton
+@Slf4j
```
StudentAlleg, line 19:
```suggestion
```

core/src/main/java/me/mykindos/betterpvp/core/command/commands/admin/FreezeCommand.java
```diff
@@ -2,6 +2,7 @@
 
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
+import lombok.extern.slf4j.Slf4j;
```
StudentAlleg, line 5:
```suggestion
```

Mykindos (approved):

PR #917, opened 2024-06-17
Fix ActionBar using null elements when non-null elements exist in queue

StudentAlleg:
Fix ActionBar components defaulting to null components when they have a higher priority, when an actual component exists that does not have a null value somewhere in the queue.

### DIFF
--- a/core/src/main/java/me/mykindos/betterpvp/core/utilities/model/display/ActionBar.java
+++ b/core/src/main/java/me/mykindos/betterpvp/core/utilities/model/display/ActionBar.java
@@ -100,7 +100,6 @@ public class ActionBar {
                 advComponent = display.getProvider().apply(gamer);
             }
 
-            assert (advComponent != null);
             // At this point, the `component` will not be null because we know that there is at least one element in the queue
             if (display instanceof TimedComponent timed) {
                 timed.startTime();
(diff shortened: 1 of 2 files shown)


core/src/main/java/me/mykindos/betterpvp/core/utilities/model/display/ActionBar.java
```diff
@@ -77,15 +80,32 @@ private Component nextComponent(Gamer gamer) {
...
+                    //there is not a valid component, return null
+                    return null;
+                }
+
+                //update display/advcomponent
+                display = pair.getRight();
+                advComponent = display.getProvider().apply(gamer);
+            }
+
+            assert (advComponent != null);
```
Mykindos, line 103: It's best not to call assert in production code, I think we can just remove this and it will still behave fine?

Mykindos (changes requested):

core/src/main/java/me/mykindos/betterpvp/core/utilities/model/display/ActionBar.java
braulio-dev: Use guava preconditions

StudentAlleg:
```suggestion
```

PR #935, opened 2024-06-25
Woodcutting - Tree Feller Improvements

8SIXSect:
## Describe your changes
- Fixed the really ambiguous cooldown message
- Made it so **Tree Feller** will work (horizontally) on the initial block that is chopped

### DIFF
--- a/progression/src/main/java/me/mykindos/betterpvp/progression/profession/skill/woodcutting/TreeFellerSkill.java
+++ b/progression/src/main/java/me/mykindos/betterpvp/progression/profession/skill/woodcutting/TreeFellerSkill.java
@@ -73,7 +73,7 @@ public class TreeFellerSkill extends WoodcuttingProgressionSkill implements Cool
 
     public void whenPlayerCantUseSkill(Player player) {
         UtilMessage.simpleMessage(player, getProgressionTree(),
-                "You must wait <alt>" + cooldownManager.getAbilityRecharge(player, getName()).getRemaining() + "</alt> seconds");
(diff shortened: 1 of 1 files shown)


clans/src/main/java/me/mykindos/betterpvp/clans/progression/perks/TreeFeller.java
```diff
@@ -83,17 +84,23 @@ public void fellTree(Clan playerClan, Block block, PlayerChopLogEvent event, boo
...
         event.setAmountChopped(event.getAmountChopped() + 1);
 
+        ArrayList<Integer> yValues = new ArrayList<>();
+        if (initialBlock) yValues.add(0);
+        yValues.add(1);
+
         for (int x = -1; x <= 1; x++) {
             for (int z = -1; z <= 1; z++) {
-                Block targetBlock = block.getRelative(x, 1, z);
+                for (int y : yValues) {
```
Mykindos, line 93: for (int y = 0; y < 1; y++) would suffice, rather than making the list just to add an int to it

Mykindos (changes requested):

clans/src/main/java/me/mykindos/betterpvp/clans/progression/perks/TreeFeller.java
8SIXSect: scrapped that change. tbh it's not worth changing. But the cooldown msg was really bothering me so just that.

PR #950, opened 2024-07-06
In game Wiki

### LINKED ISSUE #928: Add in-game mini-wiki with GUIs
Add an in-game wiki that explains certain things like effects, clans (territories, claims, slime blocks etc.), champions (classes/skills).

StudentAlleg:
Add an in game wiki, that provides easy access to important information.

This is done by implementing an IWikiable (either static or level) to a class (which can then use the actual values, for things like effects, clan sizes etc), and filling in the methods with information. These are retrieved through reflection and stored in a WikiableManager. The menu is created like the settings menu and is setup to be more easily extended. Information is defined at the place where it is implemented, to make it easier to make changes when changes are made to the thing it is referencing.

Fixes #928


StudentAlleg:
This currently in a prototype stage, not much is implemented, but the general framework for core is there. The leveled wikiable is not yet implemented fully, will eventually act similar to skills.

PR #953, opened 2024-07-09
new getNearbyEntities boolean

ezrn:
## Describe your changes
Added extra boolean variable for all variations of getNearbyEntities(including player methods), that prevents the check from counting players behind walls. Also added this restriction to quite a few skills. This boolean is false by default.


ezrn:
will improve soon and only apply it to absolutely necessary abilities such as fire blast and lightning orb

PR #954, opened 2024-07-11
Update Evade Fail Sound

wetfredhyanie:


Mykindos:
just change the volume of the button click from 2.0 to 1.0

wetfredhyanie:
> just change the volume of the button click from 2.0 to 1.0

I mean I wanted to change the whole sound, cuz it just doesn't fit. UI Button Click reminds me of iris shader reset bind which is confusing and annoying. I think the evade fail sound should be more unique.

PR #956, opened 2024-07-13
Promptable Build Menu

### LINKED ISSUE #951: Add a promptable Build Menu
Create New / Modify existing build menu to accept a roleBuild. It will then prompt the user to create the build represented by the role build.
For tutorial.

### LINKED ISSUE #952: Build and Edit Build back buttons are in different places
They should be in the same place for consistency

StudentAlleg:
Add Flashing Button, which has elements for a button to change its go at a specified time. Extends Control Item.

Create a promptable build menu, that will flash on items that the user is intended to click on, until they successfully navigate to and create the build supplied to the menu. This will be used in the tutorial to teach player how to make a build.

Fixes #951
Fixes #952

- [x] Implement Separate Command for testing
- [x] Update names of isFlashing items to indicate why they are flashing
- [x] Sync Flashing Buttons with same period

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/builds/RoleBuild.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/builds/RoleBuild.java
@@ -145,4 +145,35 @@ public class RoleBuild {
         return component;
     }
 
+    /**
+     * Returns true if the role and BuildSkills are all equal
+     * @param o the object to check against
+     * @return true if this object has the same build
+     */
+    public boolean buildEquals(Object o) {
+        if (super.equals(o)) {
+            return true;
+        }
+        if (!(o instanceof RoleBuild r2)) {
+            return false;
+        }
+
+        if (this.role != r2.getRole()) {
+            return false;
+        }
+
+        for (SkillType skillType : SkillType.values()) {
+            BuildSkill r1BuildSkill = this.getBuildSkill(skillType);
+            BuildSkill r2BuildSkill = r2.getBuildSkill(skillType);
+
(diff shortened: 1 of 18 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/champions/builds/RoleBuild.java
```diff
@@ -145,4 +145,35 @@ public Component getBuildComponent() {
         return component;
     }
 
+    /**
+     * Returns true if the role and BuildSkills are all equal
+     * @param o the object to check against
+     * @return true if this object has the same build
+     */
+    public boolean buildEquals(Object o) {
```
Mykindos, line 153: Shouldn't this just be overriding equals? doesnt seem to be getting used otherwise

Mykindos (changes requested):
The flashing on the actual build screen seems a little weird

image

Seems to change existing skills in the build to and from enchanted books

StudentAlleg:
> The flashing on the actual build screen seems a little weird
> 
> image
> 
> Seems to change existing skills in the build to and from enchanted books

Intentional? It is supposed to flash if it needs attention, so it flashes if there are points in a skill that is not in the build, if there is not enough points/too many points, and if the skill should be in the build but is not.

champions/src/main/java/me/mykindos/betterpvp/champions/champions/builds/RoleBuild.java
StudentAlleg: I think what I was going to use this for I didn't end up implementing. This method explicity compares builds without regard to the player or build #, which an equals should do. But can be removed.

Mykindos (approved):

PR #975, opened 2024-07-29
Update Agility.java

TR4E:


StudentAlleg (changes requested):
It is probably better practice to do thing all on the same branch (#974), if you add more commits to that branch, they will end up in the pull request.

TR4E:
> It is probably better practice to do thing all on the same branch (#974), if you add more commits to that branch, they will end up in the pull request.

I forgot to change the variable on a line copy and paste for the UtilMessage, hense why I created a new commit.

TR4E:
Well true, I could've basically edited it into TR4E:patch-1, didn't realize that.

PR #983, opened 2024-08-04
Brute small changes

ezrn:
Whirlwind Sword:
- Decreased scaling of radius and damage to 0.5 per level down from 1.0 per level

Whirlwind Sword was outperforming every other brute ability because it required very little user decision making for a lot of value. Tuning down the numbers should keep it as a consistent option while allowing more freedom to use other more interactive sword abilities.

Flesh Hook:
- Decreased cooldown from 18 to 13 seconds, but also decreased the cooldown decrease per level from 2 to 1.
- Hit flesh hooks also apply slowness for 1 second

These changes should help flesh hook be viable at lower levels and also reward hitting it a bit more. Flesh hook was pretty useless against assassins since they could flash away as soon as they were hit. The one second of slowness should help prevent assassins from escaping while not giving the brute any additional advantage once the enemy has been pulled to them.

Block Toss:
- Increased velocity, lowered size, lowered damage, lowered cooldown

Block toss was very easy to hit but had such a long cooldown that it was only ever used once per fight. By decreasing the cooldown, increasing the velocity, and lowering the size. It means that block toss will be more of a skill shot that has less impact but also happens more often, making the skill more viable overall. 

Takedown:
- Increased size of hit detection hitbox
- Lowered vertical height
- Stopped users velocity when they hit a target
- Lowered slowness duration scaling

This will make takedown easier to hit, but slightly less rewarding when you do at higher levels. The problem with takedown is that often you would fly over people, or when you hit them you would fly past them. These changes should help mitigate it while making the skill less oppressive at high levels and more potent at lower levels. (With the no jump change, 5 seconds of no jump slowness is much more oppressive than it used to be)

Threatening Shout:
- Increased cooldown
- Lowered strength of vulnerability
- lowered vulnerability hitbox slightly.

Threatening shout had too quick of a cooldown for how strong it was, its aoe effects were also too forgiving to warrant level 3 vulnerability. These changes should help put it in line with the power levels of the other axe abilities.

Stampede: 
- Reduced duration per stack from 6 to 5

Stampede was too fast during the alpha, but the nerf it received was quite heavy and a bit too much. This duration is a middle ground between the two extremes and should hopefully make the skill more balanced on both sides.

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/axe/Takedown.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/axe/Takedown.java
@@ -113,7 +113,6 @@ public class Takedown extends Skill implements InteractSkill, CooldownSkill, Lis
 
     @Override
     public double getCooldown(int level) {
-
         return cooldown - ((level - 1) * cooldownDecreasePerLevel);
     }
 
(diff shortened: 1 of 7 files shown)


braulio-dev:
blizzard is this you?

ezrn:
> blizzard is this you?

I figured instead of posting this and then arguing in discord for an hour over why I made these changes id just put them in the pr itself

PR #990, opened 2024-08-07
Fix bows consuming an arrow when fired

### LINKED ISSUE #968: Shooting a bow as a non assassin/ranger will consume the arrow, but not fire
https://discord.com/channels/161738857136521217/1265937620995936317/1265937620995936317
The shot is correctly cancelled, but the arrow is still consumed.

StudentAlleg:
by adding back an arrow when event is cancelled
Fixes #968 

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/roles/listeners/RoleListener.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/roles/listeners/RoleListener.java
@@ -42,6 +42,7 @@ import org.bukkit.inventory.EquipmentSlot;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.potion.PotionEffect;
 
+import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


braulio-dev:
I'd rather wait for [this](https://github.com/PaperMC/Paper/pull/9949) than give players the consumable back upon firing. This is a duplication bug waiting to happen.

8SIXSect:
I think you should leave it as a feature. Maybe add a message saying that we're keeping the arrow or something. It's a punishment for being stupid.

Mykindos (approved):

Mykindos:
Approved but I agree with what Rey says, this is a dupe waiting to happen if they fix it in a future version

StudentAlleg:
> Approved but I agree with what Rey says, this is a dupe waiting to happen if they fix it in a future version
> 
> Sure its just an arrow dupe, but dupe is a dupe

Can hold off, and if it doesn't get fixed, can include it.

PR #994, opened 2024-08-10
Add PvP Timer/Protection

### LINKED ISSUE #942: Add PvP Timer
PvP Timer is a period of time (i.e. 1 hour) that new players upon entering the map, get for the first time. While in this state, they cannot take or deal damage from players, they cannot pickup items dropped from other players, and cannot enter claims/events. Players can optionally opt-out by running a command, which is prompted to them if they try to enter a claim or attack other players.

StudentAlleg:
Add protection for new users upon joining. This time is paused in safezones.
Players that have the protection status effect cannot attack or take damage from other players. They cannot enter claims that are not admin clans (fields, shops), or their own clan.
Players can disable the protection early by running /protection

Prevent players with protection from picking up items.

Fixes #942 


Mykindos (changes requested):
I like what you've done but I dont think we should stop them from entering the territory with protection

This would be a very janky user experience, there are so many claims between safezones & admin clans, some clans having small bases in the middle of a much larger claim as well. The rubber banding alone would drive someone insane

StudentAlleg:
> I like what you've done but I dont think we should stop them from entering the territory with protection
> 
> This would be a very janky user experience, there are so many claims between safezones & admin clans, some clans having small bases in the middle of a much larger claim as well. The rubber banding alone would drive someone insane

That's a fair point, based on how the map is currently laid out. I can still see how this can be used to abuse and could see it being useful in some times, so plan on putting it behind a config flag.

PR #996, opened 2024-08-11
Add cooldown on claiming disbanded territory

### LINKED ISSUE #833: There is no cooldown on unclaimed chunks
From unclaim, from disband

StudentAlleg:
Add a cooldown on claiming territory recently unclaimed due to a Clan disband.

Fixes #833

### DIFF
--- a/clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
+++ b/clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
@@ -22,6 +22,7 @@ import me.mykindos.betterpvp.core.components.clans.IClan;
 import me.mykindos.betterpvp.core.components.clans.data.ClanAlliance;
 import me.mykindos.betterpvp.core.components.clans.data.ClanEnemy;
 import me.mykindos.betterpvp.core.components.clans.data.ClanMember;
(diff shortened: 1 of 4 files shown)


clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
```diff
@@ -177,6 +184,23 @@ public Optional<Clan> getClanByChunk(Chunk chunk) {
         return getClanById(uuid);
     }
 
+    public void applyDisbandClaimCooldown(ClanTerritory clanTerritory) {
+        setClaimCooldown(clanTerritory.getWorldChunk(), (long) (claimDisbandCooldown * 1000L));
+    }
```
Mykindos, line 189: not really necessary coinsidering all it does is call another function

Mykindos (changes requested):

clans/src/main/java/me/mykindos/betterpvp/clans/clans/ClanManager.java
StudentAlleg: I'm not sure if I want to also do things regular unclaiming, or add a way to designate an area unclaimable that admins can use (for whatever reason), which is why this was organized this way. Also, so that the config was defined in clanManager instead of the event listener.

PR #1008, opened 2024-08-13
Fix hilt smash

ezrn:
## Describe your changes
- Stopped hilt smash from activating disengage, evade, riposte

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/sword/HiltSmash.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/sword/HiltSmash.java
@@ -145,7 +145,7 @@ public class HiltSmash extends Skill implements CooldownSkill, Listener, Offensi
             UtilMessage.simpleMessage(ent, getClassType().getName(), "<yellow>%s<gray> hit you with <green>%s %d<gray>.", player.getName(), getName(), level);
             championsManager.getEffects().addEffect(ent, player, EffectTypes.SLOWNESS, slowStrength, (long) (getDuration(level) * 1000));
             UtilMessage.simpleMessage(player, getClassType().getName(), "You hit <yellow>%s<gray> with <green>%s %d<gray>.", ent.getName(), getName(), level);
-            UtilDamage.doCustomDamage(new CustomDamageEvent(ent, player, null, DamageCause.ENTITY_ATTACK, getDamage(level), false, getName()));
+            UtilDamage.doCustomDamage(new CustomDamageEvent(ent, player, null, DamageCause.CUSTOM, getDamage(level), false, getName()));
             player.getWorld().playSound(player.getLocation(), Sound.ENTITY_ZOMBIE_ATTACK_WOODEN_DOOR, 1.0F, 1.2F);
         }
     }


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/sword/HiltSmash.java
```diff
@@ -145,7 +145,7 @@ private void onInteract(Player player, LivingEntity ent) {
             UtilMessage.simpleMessage(ent, getClassType().getName(), "<yellow>%s<gray> hit you with <green>%s %d<gray>.", player.getName(), getName(), level);
             championsManager.getEffects().addEffect(ent, player, EffectTypes.SLOWNESS, slowStrength, (long) (getDuration(level) * 1000));
             UtilMessage.simpleMessage(player, getClassType().getName(), "You hit <yellow>%s<gray> with <green>%s %d<gray>.", ent.getName(), getName(), level);
-            UtilDamage.doCustomDamage(new CustomDamageEvent(ent, player, null, DamageCause.ENTITY_ATTACK, getDamage(level), false, getName()));
+            UtilDamage.doCustomDamage(new CustomDamageEvent(ent, player, null, DamageCause.CUSTOM, getDamage(level), false, getName()));
```
Mykindos, line 148: This will also stop things like deflection from reducing the damage of hilt smash, we might need to cook something else up or just leave as is

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/knight/sword/HiltSmash.java
ezrn: I feel like abilities shouldnt be resisted by stuff like that no? other solution could be adding an "ability" tag to something idk

StudentAlleg: > This will also stop things like deflection from reducing the damage of hilt smash, we might need to cook something else up or just leave as is

Deflection should reduce all damage, no?

Mykindos: > > This will also stop things like deflection from reducing the damage of hilt smash, we might need to cook something else up or just leave as is
> 
> Deflection should reduce all damage, no?

Just melee type attacks, which when you think about how a hilt smash would work in real life I think it qualifies

StudentAlleg: I mean, hilt smash doesn't have damage delay, and is instant cast (as opposed to a prepare). It's not really a melee attack. I honestly think deflection should just work for all damage, that's how I thought it worked actually tbh

ezrn: Can add a specific line to riposte, disengage, and evade to make it so hilt smash wont activate them but that seems like a bit of a cop out

PR #1028, opened 2024-08-16
Created custom weakness effect, gave it to power chop and intimindation

ezrn:
## Describe your changes
- Weakness decreases attack damage by 1.0
- Power chop gets it for 2+0.5 seconds per level
- Intimidation gets it while the enemy is looking at the brute


Mykindos:
many conflicts

PR #1034, opened 2024-08-18
Update SkillListener.java

TR4E:
## Describe your changes
Changing getClassType() to "Skill" makes more sense, considering things like "Cannot use <skill_name> while in liquid." is not a ClassType type of restriction, it's a overall "Skill".


braulio-dev (changes requested):
While you're at it, make sure the feedback messages actually use our MiniMessage tags.

- `<alt></alt>` instead of `<green>`
- `<alt2></alt2>` instead of `<yellow>`

ezrn:
come to think of it, why is it "liquid" when it only applies to water

Mykindos:
This is in a lot more places than just this, e.g. Deblink also uses the class name.

Happy with how it is for now

PR #1043, opened 2024-08-25
Woodcutting: Enchanted Lumberfall

8SIXSect:
## Describe your changes
- Everytime a player uses **Tree Feller**, they will now have a chance to get special items (like how _Fishing_ does it but without legendaries).
- What happens is one of the leaves gets removed, a small particle is spawned, and the random special item (like a rune, diamond block, stick, etc.) is placed in the player's inventory.
- But this perk also allows the player to double their drops like how _Feeling Lucky_ works
- Note: I gave the skill 250 levels but the woodcutting perk menu is really cluttered and I'll fix it eventually


progression/src/main/java/me/mykindos/betterpvp/progression/profession/woodcutting/WoodcuttingLoadConfig.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.bukkit.Material
+import org.bukkit.configuration.ConfigurationSection
+import java.util.*
+
+
+/**
+ * - Loads the YAML configuration for the Woodcutting Profession & returns the loaded data
+ * - This function will cause side effects by logging messages to the console
+ */
+fun loadConfigForWoodcutting(config: ExtendedYamlConfiguration,
```
Mykindos, line 15: Any particular reason you did this in kotlin? we have plenty of examples in java of loading config this way

Mykindos (changes requested):

ezrn:
Should it put it in the players inventory? or should it drop the special item

progression/src/main/java/me/mykindos/betterpvp/progression/profession/woodcutting/WoodcuttingLoadConfig.kt
8SIXSect: I thought the code would be more concise if put in Kotlin but I can change it to Java if want me to

8SIXSect:
> Should it put it in the players inventory? or should it drop the special item

Fishing drops the special item but fishing also doesn't have to account for obstructions being in the way. When you fish, it's in an open area with few blocks preventing you from getting the item. Even when dropping it close to the trunk of the tree, trying to find the special item felt tedious because the terrain and leaves get in the way

braulio-dev:
Actual chad for using Kotlin, you have my stamp of approval

progression/src/main/java/me/mykindos/betterpvp/progression/profession/woodcutting/WoodcuttingLoadConfig.kt
Mykindos: To keep it consistent with the other 99.9% of the project, its probably best to keep it in java

Mykindos (approved):

PR #1045, opened 2024-08-26
Replace Fury with Adrenaline

ezrn:
## Describe your changes
- Fury was just better overwhelm and therefore made no sense on brute
- Adrenaline gives you speed 1 when below a certain health threshold, and speed 2 below an even lower threshold


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/passives/Adrenaline.java
```diff
@@ -0,0 +1,122 @@
...
+                } else {
+                    Integer currentSpeedLevel = playerSpeedLevels.get(player);
+                    if (currentSpeedLevel != null && currentSpeedLevel > 0) {
+                        championsManager.getEffects().removeEffect(player, EffectTypes.SPEED);
+                        playerSpeedLevels.put(player, 0);
+                    }
+                }
+            }
+        }
+    }
```
Mykindos, line 103: A few changes needed

- In addEffect, add the name of the effect (e.g. https://github.com/Mykindos/BetterPvP/blob/c5e8793925eea427c7a8c679955268714ff77fbe/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/passives/Bloodlust.java#L106-L107)

- In removeEffect, add the same name as the last argument, this way we only remove speed provided by the passive and not speed that might come from other buffs

- To keep this as balanced as possible, just give speed for 100ms instead of 500, we dont need it running longer than intended if a player regenerates above the threshold

- I dont think this needs to run every tick, maybe every 2 ticks (can increase the duration to 150 if worried about it expiring and causing FoV issues)

- You aren't actually using the value in the playerSpeedLevels map, suggest just making this a List (either of WeakReference<Player> or just UUID)

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/passives/Adrenaline.java
```diff
@@ -0,0 +1,121 @@
...
+@Singleton
+@BPvPListener
+public class Adrenaline extends Skill implements PassiveSkill, Listener, BuffSkill, OffensiveSkill {
+
+    private double speedOneHealth;
+    private double speedOneHealthIncreasePerLevel;
+    private double speedTwoHealth;
+    private double speedTwoHealthIncreasePerLevel;
+
+    private final Set<Player> trackedPlayers = Collections.newSetFromMap(new WeakHashMap<>());
```
Mykindos, line 42: nice

Mykindos (approved):

PR #1049, opened 2024-09-02
Add LightManager, for cosmetics

StudentAlleg:
Add a way to have players give off visible light. This is done using the Light block and block updates (so is all client side). The effect is similar to optifine's lighting when you hold a torch (but does not require optifine).


StudentAlleg:
Remove light source on logout

StudentAlleg:
Re-targeting on feat/cosmetics

PR #1067, opened 2024-09-28
Add a way to view all special items

### LINKED ISSUE #1052: Add a way to view special items in game
I.E. view what a legend or rune does, without having to obtain one. Lets players have a viewable goal, and be able to understand what is going on when they face these weapons.

StudentAlleg:
Add a way to view special items with a command.
Prototype 1 shows all items, except tools, by using BPvPItem#getItemStack and ItemHandler#UpdateName, sorted by identifier.
Does not show ranges (i. e. runes/hyper), does not group runes/legends.
Only 2 rows of viewing.
A somewhat quick and dirty implementation.

Notes:
This is something very much implementable, just depends on the amount of time that should be spent implementing. Prototype 1 is better than nothing, and gives players more information, but could be improved,

Fixes #1052 


StudentAlleg:
Current Status:
Now sorts Legends and Runes to be grouped together, in the beginning. Hyper Axe shows stat variation, and Runes are being worked on.
Potential Issues:
It is currently not a singleton, so is loaded everytime the player runs the command. The items are the actual items, which means potential for duping.

PR #1073, opened 2024-09-30
Ranger Misc changes and Clean Up

ezrn:


champions/src/main/java/me/mykindos/betterpvp/champions/listeners/ArrowListener.java
```diff
@@ -77,6 +83,7 @@ public void onArrowDamage(DamageEvent event) {
                 float dmgMultiplier = arrows.get(arrow);
                 event.setDamage(event.getDamage() * dmgMultiplier);
                 arrows.remove(arrow);
+                arrow.remove();
```
braulio-dev, line 86: Are you sure this won't cause an exception? The arrow might be removed by the game upon hitting an entity, so this may cause an exception by the entity not being loaded. Just double check this

champions/src/main/java/me/mykindos/betterpvp/champions/listeners/ArrowListener.java
```diff
@@ -99,7 +106,7 @@ public void onArrowHit(ProjectileHitEvent event) {
             UtilServer.runTaskLater(champions, () -> {
                 arrow.remove();
                 arrows.remove(arrow);
-            }, 5L);
+            }, 40L);
```
braulio-dev, line 109: This might cause players to be able to pick arrows up if they spam it, can you confirm this?

champions/src/main/java/me/mykindos/betterpvp/champions/listeners/ArrowListener.java
```diff
@@ -48,8 +49,13 @@ public void onShootBow(EntityShootBowEvent event) {
         if (event.getProjectile() instanceof Arrow arrow) {
             if (event.getEntity() instanceof Player player) {
                 arrow.setMetadata("ShotWith", new FixedMetadataValue(champions, player.getInventory().getItemInMainHand().getType().name()));
+
+
+                Vector direction = player.getLocation().getDirection().normalize();
+                arrow.setVelocity(direction.multiply(event.getForce()));
+
```
braulio-dev, line 56: Why?

braulio-dev (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/listeners/ArrowListener.java
ezrn: can check if the arrow still exists before removing it, but from my testing no error messages happened

champions/src/main/java/me/mykindos/betterpvp/champions/listeners/ArrowListener.java
ezrn: The two reasons behind this change are: 
- Allow enemies to pick up missed arrows by the ranger
- Allow the ranger to better gauge where their arrow lands so they can aim the next shot more accurately

champions/src/main/java/me/mykindos/betterpvp/champions/listeners/ArrowListener.java
ezrn: This removes the randomness from arrows, which from my testing was why arrows In alpha sometimes felt strange

champions/src/main/java/me/mykindos/betterpvp/champions/listeners/ArrowListener.java
braulio-dev: Just add the piece of code to make sure

champions/src/main/java/me/mykindos/betterpvp/champions/listeners/ArrowListener.java
braulio-dev: This will have to be taken into account for in Champions but fine for now.

champions/src/main/java/me/mykindos/betterpvp/champions/listeners/ArrowListener.java
braulio-dev: AFAIK it's also because of the arrow spawn location. Mineplex made it spawn at your bow, whereas vanilla adds the randomness I think.  

In any case, you're right about the velocity.
image

Mykindos (approved):

PR #1093, opened 2024-10-11
Change Durability of Swords/Axes/Bows/Armour

RockyMichael:
## Describe your changes

Old Values (In Durability):

Basic (Stone): 400
Standard (Iron): 750
Booster (Gold):  1000
Power: (Diamond): 1000
Ancient (Netherite) 1250

Bows: 750
Crossbows: 1000

All Armour: 750

New Values:
Basic (Stone): 200
Standard (Iron): 400
Booster (Gold):  700
Power: (Diamond): 700
Ancient (Netherite) 1000

Bows: 400
Crossbows: 700

All Armour: 400

All armour and weapons are currently too durable and allow for too much usage and value to come from a single kit. The purpose of this is to make kits more valuable by providing a means for kits to be effectively removed from the economy. This will also require clans to be more conscious of their resources and kit usage.


Mykindos (changes requested):
Editing existing .sql files is no longer accepted, if you want to make this change youl need to make a new file e.g. V5__Modify_item_durabilities.sql and create update queries

PR #1116, opened 2024-10-29
Update Default Configs

StudentAlleg:
See also https://github.com/Mykindos/BetterPvP-Private/pull/37

### DIFF
--- a/clans/src/main/resources/configs/config.yml
+++ b/clans/src/main/resources/configs/config.yml
@@ -175,13 +175,13 @@ command:
         requiredRank: PLAYER
       deposit:
         enabled: true
-        requiredRank: PLAYER
+        requiredRank: ADMIN
     withdraw:
       enabled: true
       requiredRank: PLAYER
     deposit:
       enabled: true
-      requiredRank: PLAYER
+      requiredRank: ADMIN
     help:
       enabled: true
       requiredRank: PLAYER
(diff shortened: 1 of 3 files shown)


clans/src/main/resources/configs/config.yml
```diff
@@ -160,28 +169,28 @@ command:
...
-      requiredRank: ADMIN
+      requiredRank: PLAYER
       withdraw:
         enabled: true
-        requiredRank: ADMIN
+        requiredRank: PLAYER
       deposit:
         enabled: true
-        requiredRank: ADMIN
+        requiredRank: PLAYER
```
Mykindos, line 178: currently still admin only, only purpose for the bank is to store gold to collect from sold auctions

clans/src/main/resources/configs/config.yml
```diff
@@ -160,28 +169,28 @@ command:
...
     withdraw:
       enabled: true
-      requiredRank: ADMIN
+      requiredRank: PLAYER
     deposit:
       enabled: true
-      requiredRank: ADMIN
+      requiredRank: PLAYER
```
Mykindos, line 184: same

clans/src/main/resources/configs/config.yml
```diff
@@ -208,13 +217,31 @@ command:
...
     tutorial:
       enabled: true
-      requiredRank: ADMIN
+      requiredRank: PLAYER
     vault:
+      enabled: true
+      requiredRank: PLAYER
```
Mykindos, line 226: admin, gets accessed via clan core

Mykindos (changes requested):

clans/src/main/resources/configs/config.yml
StudentAlleg: gotcha

Issue #1123, opened 2024-10-30
Chopping down a default spawned oak tree gives a compacted log


StudentAlleg:
@8SIXSect 
On chopping a naturally generating oak tree, it gives the player a compacted log. Might have something to do with the custom model data being 0 for the specific item.

PR #1127, opened 2024-10-30
Swapped OAK_WOOD for CRIMSON_HYPHAE

8SIXSect:
## Describe your changes
- Let me know if I need to make *another* migration instead of changing the current ones
- Also, the reason for changing the wood type is because if you *just* change the model data on the Oak Wood from `0` -> `1` in the migrations, it doesn't fix the issue. So this fix should be a guaranteed fix

#1123 


StudentAlleg:
You will need to have a migration

Mykindos (approved):

Issue #1137, opened 2024-10-31
Create a vanish command 


StudentAlleg:
for admins to easily apply vanish to themselves

PR #1172, opened 2024-11-03
Vanish command

ItzKiba:
## Describe your changes
* Added vanish command for admins

https://github.com/Mykindos/BetterPvP/issues/1137


StudentAlleg (changes requested):
We shouldn't need to have a linkedlist of vanished players. If we were to do it this way, it should be a weakhashmap (players) or a hashmap(UUID). (this is generally how we do it for other things, like skills)

It should be just fine to check if the player was vanished (EffectManager#hasEffect) as the way to check whether to remove or apply vanish.

core/src/main/java/me/mykindos/betterpvp/core/command/commands/admin/AdminVanishCommand.java
```diff
@@ -0,0 +1,60 @@
...
+            UtilMessage.message(player, "Vanish", UtilMessage.deserialize("<red>You are no longer vanished.</red>"));
+        } else { // Not vanished
+            vanished.add(player.getUniqueId());
+            effectManager.addEffect(player, EffectTypes.VANISH, "Vanish", 1, 99999999L, true);
+            UtilMessage.message(player, "Vanish", UtilMessage.deserialize("<green>You are now vanished.</green>"));
+        }
+    }
+
+    @Override
+    public Rank getRequiredRank() {
```
StudentAlleg, line 57: This should be done in config (which defaults to the command being admin only), though it's not the worst. Just makes it slightly more difficult to change that if we want to in the future. We really shouldn't be overriding it in other places (though we do do that)

StudentAlleg (approved):

core/src/main/java/me/mykindos/betterpvp/core/command/commands/admin/AdminVanishCommand.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Override
+    public void execute(Player player, Client client, String... args) {
+        if (vanished.contains(player.getUniqueId())) { // Is already vanished
+            vanished.remove(player.getUniqueId());
+            effectManager.removeEffect(player, EffectTypes.VANISH);
+            UtilMessage.message(player, "Vanish", UtilMessage.deserialize("<red>You are no longer vanished.</red>"));
+        } else { // Not vanished
+            vanished.add(player.getUniqueId());
+            effectManager.addEffect(player, EffectTypes.VANISH, "Vanish", 1, 99999999L, true);
```
Mykindos, line 48: I think it would be better to call addEffect(LivingEntity target, LivingEntity applier, EffectType type, String name, int level, long length, boolean overwrite, boolean permanent)

Set duration to like 100L and make both booleans true

core/src/main/java/me/mykindos/betterpvp/core/command/commands/admin/AdminVanishCommand.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    public String getDescription() {
+        return "Become invisible and removes you from the tab list and auto-completions.";
+    }
+
+    @Override
+    public void execute(Player player, Client client, String... args) {
+        if (vanished.contains(player.getUniqueId())) { // Is already vanished
+            vanished.remove(player.getUniqueId());
+            effectManager.removeEffect(player, EffectTypes.VANISH);
```
Mykindos, line 44: Specify the name here also, removeEffect should take an additional argument for 'Vanish'
This way we dont remove every version of vanish a player has applied

Mykindos (changes requested):

Mykindos (approved):

PR #1273, opened 2024-11-16
Improve communication of player safe and unsafe status

### LINKED ISSUE #1231: Safe/Unsafe logging needs to be communicated better
Maybe with a logout command? The problem is, it's not easy to put this into the flow of someone logging out.

Could also notify players anytime they are unsafe.

StudentAlleg:
Improve communication of player safe/unsafe status on item pickup, combat, and territory change

Add CustomPlayerItemEvent, that tracks when a player gains or loses an item to simplify tracking this

Fixes #1231 


braulio-dev:
I'm just gonna close this and come back to it later once we've decided on what to do with UIs. I'd like to see something done with overlays using unicodes in the bossbar, or having custom indicators in the scoreboard or actionbar. @expires

PR #1298, opened 2024-11-19
Remove bank withdrawl restrictions

### LINKED ISSUE #1278: Remove restriction of withdrawing from clan bank if clan is almost pillaged
Does not make sense with the currently function of the bank

StudentAlleg:
Drop clan bank on core destroy

Fixes #1278


Mykindos:
If you remove the restriction, the clan just takes all the money out and will leave it on their player instead or hide it within a claim never to be found

StudentAlleg:
I mean, you can already do that. The bank as is currently isn't really functioning, you can only get money in it from auctions, so it's likely they already have significant amounts of money outside of it. It's just an arbitrary restriction with the current state of the bank.

PR #1322, opened 2024-11-25
ChatGPTed disable Aegis on collision mechanic

Jar-if:
Idk if it works but I figured you could review the code bc u know what ur doing and I got bored after my exam

## Describe your changes
Applies maximum level shield smash mechanic upon collision with a player, while disabling movement for [3] seconds

- [ ] I have tested my changes? No I have not


Mykindos:
Does not compile

PR #1362, opened 2024-12-07
Significant Improvements to Siphon

8SIXSect:
## Describe your changes
- Before, Siphon was an inconsistent ability and was pure RNG.
- Now, it still has an RNG aspect (gaining health) but the Speed component of the ability is more consistent
- Made great improvements to the logic of the ability
- The ability still does the same stuff, just now it does it consistently

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/passives/Siphon.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/passives/Siphon.java
@@ -4,6 +4,8 @@ import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import me.mykindos.betterpvp.champions.Champions;
 import me.mykindos.betterpvp.champions.champions.ChampionsManager;
+import me.mykindos.betterpvp.champions.champions.builds.menus.events.SkillDequipEvent;
+import me.mykindos.betterpvp.champions.champions.builds.menus.events.SkillEquipEvent;
 import me.mykindos.betterpvp.champions.champions.skills.Skill;
 import me.mykindos.betterpvp.champions.champions.skills.types.BuffSkill;
 import me.mykindos.betterpvp.champions.champions.skills.types.HealthSkill;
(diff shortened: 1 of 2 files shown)


8SIXSect:
Updated description

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/passives/Siphon.java
```diff
@@ -70,52 +98,151 @@ public double getEnergySiphoned(int level) {
...
+    @UpdateEvent (delay = 500)
+    public void removeActives() {
+        playerUUIDToEnemyDataMap.keySet().forEach(playerUUID -> {
+            if (!active.contains(playerUUID)) playerUUIDToEnemyDataMap.remove(playerUUID);
+        });
+    }
+
+    @UpdateEvent(delay = SIPHON_UPDATE_DELAY)
+    public void monitorActives() {
+        active.forEach(playerUUID -> {
```
Mykindos, line 158: I feel like you wouldn't a seperate active hashmap, just do it all within the playerUUIDToEnemyDataMap (maybe rename it to something less bad, like just siphonData or something

Mykindos: Then iterate it with a standard iterator and remove entries if they dont have the skill etc

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/passives/Siphon.java
8SIXSect: I was able to condense the logic into a singular HashMap but do I need two UpdateEvents or is the way I'm doing things redundant?

Mykindos: Just need one but use an iterator instead of forEach

Mykindos (approved):

PR #1375, opened 2024-12-11
Refactored Arrow Trail logic

8SIXSect:
## Describe your changes
- Greatly simplified the logic for arrow trails and made it uniform across bow skills (even special ones that don't use the `arrows` list)
- Added a cool arrow trial for Kinetics
- Made the Vitality Spores particle slightly more visible
- Added a particle to Barbed Arrows (when you hit a player) that is very similar to Vitality Spores's
- Also removed Barbed Arrows's arrow trail (as it was the same as Heavy Arrows's)

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/Skill.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/Skill.java
@@ -406,12 +406,6 @@ public abstract class Skill implements IChampionsSkill {
         return hasSkill(player) && SkillWeapons.isHolding(player, getType());
     }
 
-
-    public static <T extends Projectile> void updateParticleForArrowTrail(Particle particle, Iterator<T> it,
-                                                                          int receiversRadius, boolean checkIfInBlock) {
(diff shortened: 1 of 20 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/Skill.java
```diff
@@ -392,4 +406,26 @@ protected int getLevel(Player player) {
...
+                it.remove();
+            } else if (checkIfInBlock && arrowOrProjectile instanceof Arrow && ((Arrow) arrowOrProjectile).isInBlock()) {
+                it.remove();
+            } else {
+                Location location = arrowOrProjectile.getLocation().add(vector);
+                ParticleBuilder particleBuilder = builderFunction.apply(location);
+                particleBuilder.spawn();
+            }
+        }
+    }
```
Mykindos, line 430: Feel like you could move this to the PrepareArrowSkill subclass and then handle the update events for all of them within that as well

Mykindos (changes requested):

PR #1381, opened 2024-12-13
Fixed tri shot damage and made it compatible with hunter's thrill

8SIXSect:
## Describe your changes
🐐

#trii shot bug issue somewhere idk couldnt find

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/bow/TriShot.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/bow/TriShot.java
@@ -34,6 +34,7 @@ import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Player;
 import org.bukkit.entity.Trident;
 import org.bukkit.event.EventHandler;
+import org.bukkit.event.EventPriority;
 import org.bukkit.event.block.Action;
 import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.event.entity.PlayerDeathEvent;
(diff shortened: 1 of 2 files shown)


champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/bow/TriShot.java
```diff
@@ -244,10 +245,11 @@ public void displayTrail(Location location) {
...
     public void removeTridents(ProjectileHitEvent event) {
-        if (event.getEntity() instanceof Trident trident) {
-            trident.remove();
-            tridents.remove(trident);
-        }
+        if (event.isCancelled()) return;
+        if (!(event.getEntity() instanceof Trident trident)) return;
+
+        // When an entity is hit, that is handled elsewhere
+        if (event.getHitEntity() == null) trident.remove();
```
braulio-dev, line 252: This does the opposite of the comment? Or am I missing something?

braulio-dev (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/bow/TriShot.java
8SIXSect: Ok so the event listens for a projectile hit event; specifically, for tridents. This event handler pretty much handles any time that an entity (like cow, pig, etc.) is NOT hit. So the purpose of this event is to remove a trident when it gets stuck in a block.

The place I was referring to with "that is handled elsewhere" is [here](https://github.com/Mykindos/BetterPvP/blob/fe88f1e795f9a241852e8ca02429f6ce5bb7bb5c/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/ranger/bow/TriShot.java#L270)

PR #1400, opened 2024-12-24
Update default config

StudentAlleg:
Pushed in project

### DIFF
--- a/core/src/main/resources/configs/config.yml
+++ b/core/src/main/resources/configs/config.yml
@@ -272,7 +272,41 @@ command:
     requiredRank: PLAYER
   silencechat:
     enabled: true
+    requiredRank: MODERATOR
+  punishmenthistory:
+    enabled: true
+    requiredRank: PLAYER
+  unloadpack:
+    enabled: true
     requiredRank: ADMIN
+  legacypunish:
+    enabled: true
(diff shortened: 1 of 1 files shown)


core/src/main/resources/configs/config.yml
```diff
@@ -271,8 +271,42 @@ command:
     enabled: true
     requiredRank: PLAYER
   silencechat:
+    enabled: true
+    requiredRank: MODERATOR
+  punishmenthistory:
+    enabled: true
+    requiredRank: PLAYER
```
Mykindos, line 278: you can use this to see other players, so maybe best to keep it as a mod command

Mykindos (changes requested):

core/src/main/resources/configs/config.yml
StudentAlleg: https://github.com/Mykindos/BetterPvP/blob/6bef19edb84a3c2601134b2a55ef6888c16009d2/core/src/main/java/me/mykindos/betterpvp/core/client/punishments/commands/PlayerPunishmentHistoryCommand.java#L45-L67
only helpers and above can see other player's punish history

Mykindos (approved):

PR #1429, opened 2024-12-30
Change chat filter to filter out messages rather than block

BillyDotWS:
## Describe your changes
- Use the ChatModule's getFilteredMessage method to replace any filtered words
- Remove the event cancellation if filter is detected
- Remove the bypass for admins.


Mykindos:
Used this to guide property implementation

PR #1531, opened 2025-04-11
Updated Class Selection Menu to Feature Role Passives

8SIXSect:
## Describe your changes
- Added a toggle feature to the class selection menu so players have the option to see role passives
- mykindos please add more role passives

## Demo
https://github.com/user-attachments/assets/0afebc44-9db4-4363-a578-45a8c2b6acfb

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/roles/listeners/roleeffects/AssassinListener.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/roles/listeners/roleeffects/AssassinListener.java
@@ -32,17 +32,14 @@ import java.util.ArrayList;
 public class AssassinListener implements Listener, ConfigAccessor {
 
     // <editor-fold defaultstate="collapsed" desc="Config">
-    private boolean surgicalPrecisionIsEnabled;
-    private String surgicalPrecisionName;
-    private boolean surgicalPrecisionIsBuff;
+    private boolean meleeDealsNoKnockbackIsEnabled;
+    private boolean meleeDealsNoKnockbackIsBuff;
 
-    private boolean blurIsEnabled;
-    private String blurName;
-    private boolean blurIsBuff;
+    private boolean speedBuffIsEnabled;
+    private boolean speedBuffIsBuff;
 
-    private boolean speedlockIsEnabled;
-    private String speedlockName;
-    private boolean speedlockIsBuff;
+    private boolean noKnockbackReceivedWhenSlowedIsEnabled;
+    private boolean noKnockbackReceivedWhenSlowedIsBuff;
     // </editor-fold>
 
     private final RoleManager roleManager;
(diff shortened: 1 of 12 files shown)


champions/src/main/resources/configs/config.yml
```diff
@@ -1,7 +1,17 @@
...
+      name: "Surgical Precision"
+      isBuff: true
+    blur:
+      enabled: true
+      name: "Blur"
+      isBuff: true
+    speedlock:
+      enabled: true
+      name: "Speedlock"
+      isBuff: false
```
Mykindos, line 14: Instead of giving these passives bad names, maybe just display them like

\+ Effect 1
\+ Effect 2
\- Effect 3

Mykindos (changes requested):

champions/src/main/resources/configs/config.yml
braulio-dev: I agree, we don't need any more confusing names

StudentAlleg:
I'm not sure a toggle is necessary (from a UI standpoint). There are too many things that have a setting/toggle that don't really need it. (others may disagree)

champions/src/main/resources/configs/config.yml
StudentAlleg: I can see your point,, I actually thing it makes more sense with the names though.

braulio-dev: There's no need for a name if there isn't a need for anybody to ever reference the passive. If there's nothing to choose from then they just exist, there's no need to make a callout like "this person has surgical precision on". It just adds more things that the player has to remember. 

Adding a name does not mean it's "simpler", sometimes not doing anything at all is the best way

8SIXSect:
here is what no names looks like

Mykindos:
>  here is what no names looks like

You could even remove the labels effect 1 effect 2 etc

Mykindos:
> > >  here is what no names looks like
> > 
> > 
> > You could even remove the labels effect 1 effect 2 etc
> 
>  with no labels
> 
> Or maybe this one is better: 
> 
> Actually i like this one the best with better descriptions: 

Much better

champions/src/main/java/me/mykindos/betterpvp/champions/champions/roles/listeners/roleeffects/AssassinListener.java
```diff
@@ -0,0 +1,138 @@
...
+        this.surgicalPrecisionName = config.getOrSaveString("class.assassin.surgicalPrecision.name", "Surgical Precision");
+        this.surgicalPrecisionIsBuff = config.getOrSaveBoolean("class.assassin.surgicalPrecision.isBuff", true);
+
+        this.blurIsEnabled = config.getOrSaveBoolean("class.assassin.blur.enabled", true);
+        this.blurName = config.getOrSaveString("class.assassin.blur.name", "Blur");
+        this.blurIsBuff = config.getOrSaveBoolean("class.assassin.blur.isBuff", true);
+
+        this.speedlockIsEnabled = config.getOrSaveBoolean("class.assassin.speedlock.enabled", true);
+        this.speedlockName = config.getOrSaveString("class.assassin.speedlock.name", "Speedlock");
+        this.speedlockIsBuff = config.getOrSaveBoolean("class.assassin.speedlock.isBuff", false);
```
Mykindos, line 135: give these generic names associated with what the buff does, e.g. speed-buff-passive

champions/src/main/resources/configs/config.yml
Mykindos, line 14: same

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/champions/roles/listeners/roleeffects/AssassinListener.java
8SIXSect: done

Mykindos (approved):

PR #1575, opened 2025-05-03
Prevent skills that only work on allies affect enemies and vice versa

### LINKED ISSUE #1571: Champions: Some buffs apply to enemies
From _Salt in game, buffs apply to enemies, specifically arctic armor and defensive aura
- [ ] Confirmed

StudentAlleg:
FetchNearbyEntityEvent was not being handled correctly by TeamDamageListener, it only set if friendly, it also needs to set if enemy.

Functions that used FetchNearbyEntityEvent were not checking if the entities being fetched actually matched the desired property.

Fixes #1571


StudentAlleg:
I think I messed up my boolean logic will need to fix

StudentAlleg:
fixed

Mykindos (approved):

PR #1580, opened 2025-05-04
Correctly cancel Smoke Bomb on flag pickup

### LINKED ISSUE #1567: Correctly cancel smoke bomb on picking up the flag
Currently, smoke bomb's vanish effect is canceled on picking up the flag. It should instead actual trigger the skill's end function and remove it from the tracked players.
Either, get the skill from ChampionsSkillManager and use one of those methods (requires correctly getting that into the right place). Or having a new event like Skill interrupt event, which is probably more appropriate for a skill refactor.

StudentAlleg:
Fixes #1567

### DIFF
--- a/game/src/main/java/me/mykindos/betterpvp/game/impl/ctf/model/FlagPlayerHandler.java
+++ b/game/src/main/java/me/mykindos/betterpvp/game/impl/ctf/model/FlagPlayerHandler.java
@@ -2,9 +2,6 @@ package me.mykindos.betterpvp.game.impl.ctf.model;
 
 import java.util.HashMap;
 import java.util.Map;
-import me.mykindos.betterpvp.champions.Champions;
-import me.mykindos.betterpvp.champions.champions.skills.ChampionsSkillManager;
-import me.mykindos.betterpvp.champions.champions.skills.skills.assassin.passives.SmokeBomb;
 import me.mykindos.betterpvp.core.effects.EffectManager;
 import me.mykindos.betterpvp.core.effects.EffectTypes;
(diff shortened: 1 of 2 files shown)


Mykindos (approved):

game/src/main/java/me/mykindos/betterpvp/game/impl/ctf/model/FlagPlayerHandler.java
```diff
@@ -43,7 +49,12 @@ public void pickUp(Player holder) {
 
         // Effect
         effectManager.removeEffect(holder, EffectTypes.SPEED);
-        effectManager.removeEffect(holder, EffectTypes.VANISH, "Smoke Bomb");
+
+        //Skill
+        skillManager.getObject("Smoke Bomb").ifPresent(skill -> {
+            SmokeBomb smokeBomb = (SmokeBomb) skill;
+            smokeBomb.remove(holder);
+        });
```
Mykindos, line 57: Feel like just listening to effectremoveevent inside of Smoke Bomb would be better

Mykindos (changes requested):

game/src/main/java/me/mykindos/betterpvp/game/impl/ctf/model/FlagPlayerHandler.java
StudentAlleg: that makes more sense

StudentAlleg: fixed

Mykindos (approved):

PR #1581, opened 2025-05-04
Allow certain Mineplex commands

### LINKED ISSUE #1564: Allow specific Mineplex commands in CommandListener
Also check if they are a Mineplex command (or a simple check if in set).

StudentAlleg:
Fixes #1564


Mykindos (changes requested):
This only allows the tab completion to work I think. You need to add it to the preprocess

StudentAlleg:
> This only allows the tab completion to work I think. You need to add it to the preprocess

Whoops, forgot differences between this and my changes made in brigadier. Fixed

Mykindos (approved):

PR #1593, opened 2025-05-04
Fix resilience removing flag slowness

StudentAlleg:
An interaction between Long.MAX_VALUE and resilience was causing flag slow to not effect those players

### DIFF
--- a/game/src/main/java/me/mykindos/betterpvp/game/impl/ctf/model/FlagPlayerHandler.java
+++ b/game/src/main/java/me/mykindos/betterpvp/game/impl/ctf/model/FlagPlayerHandler.java
@@ -48,7 +48,7 @@ public class FlagPlayerHandler implements HatProvider, ItemProvider, Lifecycled 
     }
 
     public void tick(Player holder) {
-        effectManager.addEffect(holder, holder,EffectTypes.SLOWNESS, "Flag", 1, 600 * 1000);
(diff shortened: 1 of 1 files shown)


game/src/main/java/me/mykindos/betterpvp/game/impl/ctf/model/FlagPlayerHandler.java
```diff
@@ -48,9 +48,7 @@ public void pickUp(Player holder) {
     }
 
     public void tick(Player holder) {
-        if (!effectManager.hasEffect(holder, EffectTypes.SLOWNESS, "Flag")) {
-            effectManager.addEffect(holder, EffectTypes.SLOWNESS, "Flag", 1, Long.MAX_VALUE);
-        }
+        effectManager.addEffect(holder, holder,EffectTypes.SLOWNESS, "Flag", 1, 600 * 1000);
```
Mykindos, line 51: Set it to a duration of 100, and set overwrite to true. That way the list doesnt get filed with duplicate effects

Mykindos (changes requested):

game/src/main/java/me/mykindos/betterpvp/game/impl/ctf/model/FlagPlayerHandler.java
StudentAlleg: yikes yep, fixed

Mykindos (approved):

PR #1595, opened 2025-05-05
Only apply the highest amplifier vanilla effect

### LINKED ISSUE #1594: Only the highest amplifier effect should tick
Currently, all effects tick. Because of this, it can lead to effects fighting each other if the user has an effect. Only one effect of the same type should be actively applied, that being the one that has the highest amplifier, then by duration.

StudentAlleg:
Only apply the highest amplifier -> highest duration vanilla effect. This fixes an issue where adrenaline and stampede would flicker between level 2 and 3. In general, now only 1 vanilla effect is applied on a player at a time.

Fixes #1594


StudentAlleg:
wrong branch

PR #1615, opened 2025-05-06
Keep armor while spectating to allow skills and energy to recharge

### LINKED ISSUE #1599: Spawn players with full charge of skills and energy
Players do not spawn with skills and energy. This is probably because they do not have a kit while spectating.

StudentAlleg:
Fixes #1599


Mykindos:
@StudentAlleg some skills can be used while in spectator mode i think, e.g. fireblast

StudentAlleg:
> @StudentAlleg some skills can be used while in spectator mode i think, e.g. fireblast

Well, you'd need an axe to use it, which they won't have (hotbar is not set, only the armor)

Mykindos (approved):

PR #1634, opened 2025-05-08
Update SmokeBomb.java

TR4E:
## Describe your changes
Changes the "Assassin> You have reappeared" to "Smoke Bomb> You have reappeared", seems better when shown the Skill Name because it refers to it

- [✅] I have performed a self-review of my code
- [✅] I have tested my changes.


Mykindos:
This would become the only skill to do this, probably not going to change it

TR4E:
> This would become the only skill to do this, probably not going to change it

Might aswell look into other skills with the same 'wrong reference' and change it to make it look better

PR #1686, opened 2025-05-23
Achievements

StudentAlleg:
WIP
Achievements:
Tied to PropertyContainers, track achievements that they make based off of stats in their container.

Basic premise is that achievements are attatched to a specific property container (i.e. Gamer). When a gamer property is updated, the achievement checks to see if that property is tracked by an achievement. If so, it sends that data, along with the non-changed data, to the achievement itself, which will handle notifying the container (i.e. player), handling any rewards etc. (this might need to track the previous value as well, depending on incrementing and things like that)

Simple achievements, for things that can only go 1 direction (like increasing playtime, kills) only require that the metric is tracked in the specified container. If it could change (like whether a player is on lunar or not), another stat that is static needs to be created.

## TODO
- [ ] GUI display
- [x] Single Increasing Achievements
- [ ] Single decreasing Achievements
- [ ] Combined Achievements
- [ ] Correct notification/reward handling (i.e. not saying completed more than once)
- [x] Loader
- [x] Config
- [ ] Config support (load some/certain achievements from config, like dungeons)


core/src/main/java/me/mykindos/betterpvp/core/client/stats/ClientStat.java
```diff
@@ -0,0 +1,66 @@
...
+
+@Getter
+public enum ClientStat implements IClientStat {
+    //clans
+    SET_CORE("Set Core", "Number of times you set your clan core"),
+    TELEPORT_CORE("Teleport to Core", "Number of times you teleported to your clan core"),
+
+    //events
+    DREADBEARD_KILLS("Kill Dreadbeard", "Number of times you killed Dreadbeard"),
+    SKELETON_KING_KILLS("Kill the Skeleton King", "Number of times you killed the Skeleton King")
```
StudentAlleg, line 15: Currently all stats are defined here in core (See also Minecraft stats, which are a more special case). Originally the idea was to have it per plugin, which is why Achievements has per-plugin loading and stats does not currently. I'm not sure if I should change stats over to that.

core/src/main/java/me/mykindos/betterpvp/core/client/stats/MinecraftStat.java
```diff
@@ -0,0 +1,187 @@
...
+import org.bukkit.Material;
+import org.bukkit.Statistic;
+import org.bukkit.entity.EntityType;
+import org.bukkit.event.player.PlayerStatisticIncrementEvent;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@Builder
+@Getter
+public class MinecraftStat implements IStat {
```
StudentAlleg, line 15: Minecraft by default collects statistics, though they are only per server/world. Handling all of the variations by manually defining them would not be feasible, this class handles both serializing and parsing the stats for generic use. Some Statistics have qualifiers, so there are a lot of variations to store and retrieve. Also added the ability for those with qualifiers to easily sum them.

core/src/main/java/me/mykindos/betterpvp/core/client/stats/StatContainer.java
```diff
@@ -0,0 +1,84 @@
...
+import me.mykindos.betterpvp.core.framework.customtypes.IMapListener;
+import me.mykindos.betterpvp.core.utilities.model.Unique;
+import org.bukkit.plugin.java.JavaPlugin;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Map;
+import java.util.UUID;
+
+@CustomLog
+public class StatContainer implements Unique, IMapListener {
```
StudentAlleg, line 17: Sort of the base of stats, this is similar to the PropertyContainer classes but not the same. Specifically, saving the period.

core/src/main/java/me/mykindos/betterpvp/core/client/stats/StatContainer.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * The current period of stat collecting
+     */
+    public static final String PERIOD = JavaPlugin.getPlugin(Core.class).getConfig().getOrSaveString("stats.period", "test");
```
StudentAlleg, line 21: Because stats are stored in the global database, we need a way to differentiate when/where they were collected. My solution is the period, so something like "2025 Beta 3". Achievements definitely need this functionality, as some achievements should be specific to certain periods.

champions/src/main/java/me/mykindos/betterpvp/champions/achievements/loader/ChampionsAchievementLoader.java
```diff
@@ -0,0 +1,45 @@
...
+    protected void loadAchievementCategories(String packageName) {
+        Reflections reflections = new Reflections(packageName);
+        Set<Class<? extends IAchievementCategory>> classes = reflections.getSubTypesOf(IAchievementCategory.class);
+        loadAll(classes);
+
+        Set<Class<?>> subCategoryClasses = reflections.getTypesAnnotatedWith(SubCategory.class);
+        loadSubCategories(subCategoryClasses);
+
+        plugin.saveConfig();
+        log.error("Loaded {} categories for Core", count).submit();
```
Mykindos: should be Champions

Mykindos (changes requested):

champions/src/main/java/me/mykindos/betterpvp/champions/achievements/impl/UseAllSkillsAchievement.java
```diff
@@ -0,0 +1,122 @@
...
+                .map(GenericStat.class::cast)
+                .map(GenericStat::getStat)
+                .map(ChampionsSkillStat.class::cast)
+                .toList();
+        List<Component> components = new ArrayList<>();
+        if (neededStats.isEmpty()) return List.of();
+        components.add(Component.text("Needed:"));
+        for (int i = 0; i < 3; i++) {
+            components.add(Component.text(Objects.requireNonNull(neededStats.get(i).getSkillName()), NamedTextColor.WHITE));
+        }
```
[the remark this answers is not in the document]
StudentAlleg: good catch

champions/src/main/java/me/mykindos/betterpvp/champions/achievements/impl/UseAllSkillsAchievement.java
```diff
@@ -0,0 +1,122 @@
...
+public class UseAllSkillsAchievement extends NSingleGoalSimpleAchievement {
+
+    @Inject
+    public UseAllSkillsAchievement(ChampionsSkillManager skillManager) {
+        super("Use All Skills", new NamespacedKey("champions", "use_all_skills"),
+                AchievementCategories.CHAMPIONS,
+                StatFilterType.ALL, 60_000L,
+                getAllSkills(skillManager));
+        //cannot register via BPvPListener as it loads before skills are loaded
+        ListenerLoader.register(JavaPlugin.getPlugin(Champions.class), this);
```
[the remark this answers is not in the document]
StudentAlleg: The loading of listeners happens before skills are loaded, so the achievement is then initialized with no skills to get (it uses the current skill list to generate the IStat's it is listening too).

PR #1756, opened 2025-07-21
Enable purification potions lol

twitching12:
## Describe your changes

- Enable purification potions


Mykindos:
This will also enable it on clans, which we dont want to do it.
I have implemented a champions specific override that will fix this next restart

PR #1794, opened 2025-10-06
Let mods start champs games

8SIXSect:
## Describe your changes
- Game command isnt in config.yml so all the perms default to ADMIN. I'm assuming it's not in the config file for a reason so I put the required rank in the source code for `GameCommand.java`.

### DIFF
--- a/game/src/main/java/me/mykindos/betterpvp/game/command/GameCommand.java
+++ b/game/src/main/java/me/mykindos/betterpvp/game/command/GameCommand.java
@@ -4,7 +4,6 @@ import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import lombok.CustomLog;
 import me.mykindos.betterpvp.core.client.Client;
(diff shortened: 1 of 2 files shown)


StudentAlleg (changes requested):
This should be done in the games config.yml under commands not hardcoded

(if it isn't in Game's config.yml it should be)

StudentAlleg:
> > This should be done in the games config.yml under commands not hardcoded
> 
> 
> 
> yeah that's what i thought too but i assumed that it wasn't put in the config.yml for a reason

Stuff that isn't in config (i.e. commands) should generally be in there, just when we added the command we forgot to also do the auto generated config stuff.

StudentAlleg (approved):

braulio-dev (changes requested):
Make all the subcommands accessible to Moderator too, I want to give them access to change the minimum required players

PR #1804, opened 2025-11-05
Postgres jooq migration

Mykindos:
## Describe your changes
- Move from MariaDB to Postgres to make use of a lot of optimisations and better partitioning
- Add partitions by realm or season to tables expected to get a lot of data
- Added a lot of new indexes and foreign key relationships
- Internally use Client ID (long) instead of UUID for joins
- Create a Server / Realm concept to allow for smaller indexes on tables such as logs
- Convert to from homemade SQL to [jOOQ](https://www.jooq.org/doc/3.21/manual/), allowing for typesafe queries and cross-dialect support should we wish for it

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -11,7 +11,6 @@ plugins {
     id("io.papermc.paperweight.userdev") version libs.versions.paperweight apply false // NMS Paper
     id("org.flywaydb.flyway") version "11.15.0" apply false // Flyway
     id("org.sonarqube") version "7.0.1.6134" apply true
-
 }
 
 java {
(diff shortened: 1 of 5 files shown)


build.gradle.kts
```diff
@@ -11,6 +11,7 @@ plugins {
     id("io.papermc.paperweight.userdev") version libs.versions.paperweight apply false // NMS Paper
     id("org.flywaydb.flyway") version "11.15.0" apply false // Flyway
     id("org.sonarqube") version "7.0.1.6134" apply true
+
```
braulio-dev: whitespace

PR #1849, opened 2026-02-10
A New Shape for the Threatening Shout

8SIXSect:
## Deep-rooted Issues
Walking sucks. Mobility is fun. Why is **Kinetics** so popular despite Ranger being severely underpowered in close-quarters combat? Because players like to move around with swiftness and creativity.

Yet, Brute's Axe abilities forces you to choose between 3 movements skills (**Seismic Slam**, **Takedown**, and **Skullsplitter**) or a purely offensive skill (**Threatening Shout**). No matter how good that outlier offensive skill is, players won't want to pick it. It's not a fun choice.

## Threatening Shout's New Place
The fix is simple: Move **Threatening Shout** from an Axe skill to a Passive A.

Here, it can still be an "active" skill like **Smoke Bomb** and **Recall**; yet, Brutes no longer have to choose between having mobility or not having mobility.

image

## Brute Gains A Finisher
Moreover, **Threatening Shout** has some new properties to it. Instead of applying _Vulnerability_ to enemies, the skill deals more damage to low health enemies. Thus, this skill is no longer in a feast-or-famine state but instead has a dedicated role.

Not to mention the potential combos that can now be pulled off with a 3rd active ability on Brute:
https://medal.tv/games/minecraft/clips/m5sGB4B4bfGLozSW7?invite=cr-MSxRd3QsMzQxNDU1NjM4

## In A Perfect World...
Unfortunately, this would leave Brute's skills in an awkward situation. **Intimidation** and **Fortify** would go from dead to extinct while **Stampede** and **Bloodlust** would be relegated to "situational" status; nevertheless, giving Brute players another tool is a step towards a new meta that ALL players can enjoy.

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/passives/ThreateningShout.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/brute/passives/ThreateningShout.java
@@ -181,7 +181,7 @@ public class ThreateningShout extends Skill implements Listener, CooldownToggleS
 
     @Override
     public SkillType getType() {
-        return SkillType.PASSIVE_B;
+        return SkillType.PASSIVE_A;
     }
 
     @Override


Mykindos:
I'm not sure how I feel about a skill shot being a passive

PR #1886, opened 2026-03-13
Fix teleport skills blocking thread execution and starving resources. Change to completable future for activate (that is used to track whether a skill was used or not)

StudentAlleg:
Teleport async wouldn't complete in time for a completable future join, which was only done so that the success of a skill usage could be more accurately tracked. Now, change that activation to a completable future and increment the stat when it is done.


StudentAlleg:
Moved to a student/archive branch. If accurately tracking usages becomes important, we can revisit this.

PR #1940, opened 2026-04-13
Assassin Passive B: Deathwish

8SIXSect:
### A boring meta
Assassin has always had lackluster passives in terms of both visual appeal and combat viability. Skills such as **Combo Attack** and **Backstab** have always been viable but are bland while **Silencing Strikes**, **Shocking Strikes**, and **Viper Strikes**\* have more interesting possibilities to them but suck pretty bad.

### Enter: Deathwish
Deathwish is a versatile skill rewarding players who thrive playing as a glass cannon; yet, also provides unique mechanics like boosted *Attack Speed* and an interesting health mechanic.

image

### Demo
https://medal.tv/games/minecraft/clips/mv3g9a8WH6rPcuri3?invite=cr-MSx3bmQsMzQxNDU1NjM4

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/data/deathwish/DeathwishData.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/assassin/data/deathwish/DeathwishData.java
@@ -1,0 +1,19 @@
+package me.mykindos.betterpvp.champions.champions.skills.skills.assassin.data.deathwish;
+
+import lombok.Data;
+import org.jetbrains.annotations.NotNull;
+
+@Data
+public class DeathwishData {
+    private @NotNull DeathwishThreshold currentThreshold = DeathwishThreshold.NONE;
+
+    /**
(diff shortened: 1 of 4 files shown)


Mykindos:
Conceptually its cool but in practice its just another major damage increase when it comes to duels. It doesn't require any thinking so I dont see how it addresses the so called boring meta

8SIXSect:
> Conceptually its cool but in practice its just another major damage increase when it comes to duels. It doesn't require any thinking so I dont see how it addresses the so called boring meta

You are correct that it requires little thinking compared to other skills. Yet it still, addresses the "boring meta" through new **fun** playstyles opened for assassin.

For instance, Assassin is not an impactful kit in teamfights. Assassins go in and get deleted or their long cooldowns take them out of the fight. With Deathwish, players can take a last stand after using **Smoke Bomb** or **Recall** and could turn the tide of a fight.

Builds will also be shaken up. For better players, using this skill at level 1 or 2 will be viable. With an extra couple skill tokens, bow skills could see a resurgence. This skill also synergizes well with lesser used sword skills like **Excessive Force** (since it boosts attack speed) or **Concussion** (since the gap in attack speeds should make it much better).

PR #1969, opened 2026-04-21
Fix seeing border for warning effect

StudentAlleg:
Increase border size so players do not see border


braulio-dev (commented):
I'd like to actually see this in game or you verify that it does work, since I'm unsure if the warning distance being the size of the border will retain the vignette behavior. It was set to `1.0` because that would put players outside the border which applies the full vignette.

Additionally, this might throw an error if `curBorder.getMaxSize()` is less than the hardcoded `100_000`, as per [these javadocs](https://jd.papermc.io/paper/26.1.2/org/bukkit/WorldBorder.html#setSize(double)).

PR #2006, opened 2026-05-06
Warlock Passthrough

### LINKED ISSUE #2002: Go through warlock and do update balance changes
Cooldowns, health for abilities

StudentAlleg:
Balance pass on warlock. Mostly adding config changes already present in code to skills .yml. Reduce cooldowns on some sword skills to be more reasonable

### DIFF
--- a/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/sword/BloodSphere.java
+++ b/champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/sword/BloodSphere.java
@@ -191,10 +191,10 @@ public class BloodSphere extends Skill implements CooldownSkill, InteractSkill, 
         this.applyTravelSpeed = getConfig("applyTravelSpeed", 6.0, Double.class);
         this.damagePerSecond = getConfig("damagePerSecond", 3.0, Double.class);
         this.damageIncreasePerLevel = getConfig("damageIncreasePerLevel", 1.0, Double.class);
-        this.regenPerSecond = getConfig("regenPerSecond", 2.0, Double.class);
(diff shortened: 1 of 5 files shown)


champions/src/main/resources/configs/skills/skills.yml
```diff
@@ -1113,8 +1113,8 @@ skills:
       rangeIncreasePerLevel: 0.5
       baseDamageIncrease: 0.33
       damageIncreasePerLevel: 0.0
-      baseHealthReduction: 8.0
```
8SIXSect, line 1116: What if you left all the Health Sacrifices for the axe abilities as is? All the abilities are pretty weak. If the healing/damage output of the kit is too great, then this can be done later, no?

champions/src/main/java/me/mykindos/betterpvp/champions/champions/skills/skills/warlock/sword/BloodSphere.java
```diff
@@ -191,10 +191,10 @@ public void loadSkillConfig() {
         this.applyTravelSpeed = getConfig("applyTravelSpeed", 6.0, Double.class);
```
8SIXSect, line 191: What if you gave a slight buff to the travel speed? Blood Sphere has always been a super weak skill but it was decently use-able during the Reybot's incident.

champions/src/main/resources/configs/skills/skills.yml
```diff
@@ -1162,22 +1164,22 @@ skills:
     grasp:
```
8SIXSect, line 1164: I like this buff. But Grasp will still probably be the worst sword skill. What if you made it so if you hit an enemy after using Grasp on them, Grasp will reactive (up to 1 time only). This would allow for combos without changing the damage/distance values of the skill.

PR #2014, opened 2026-05-07
Vengeance Bug Fixes + Visual Effects

8SIXSect:
## Describe your changes
- Made `Vengeance` description cleaner.
- Fixed `Vengeance` bugs (it was triggering on cancelled attacks; was counting every instance of damage as a "hit" instead of just looking at melee hits; had a lot of old code)
- Added feedback for player (player can now see how many hits they've taken)
- Added small bonus effect:
- - When a player hits an enemy while at max `Vengeance` stacks, they pull all nearby enemies to them (weak pull, small range).
- - This new effect synergizes well with the entire `Knight` kit as abilities like `Vanguard's Might`, `Hold Position`, and `Riposte` can now leverage their damage reduction in a meaningful way.

## Demo
https://medal.tv/games/minecraft/clips/mFekiT0DRXfqcClaa?invite=cr-MSw4MGgsMzQxNDU1NjM4

image

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -106,7 +106,7 @@ subprojects {
         archiveBaseName.set(project.name)
         archiveVersion.set("")
         archiveClassifier.set("")
-        destinationDirectory.set(file("$rootDir/build/"))
+        destinationDirectory.set(file("../../TestServer/plugins"))
         mergeServiceFiles()
         duplicatesStrategy = DuplicatesStrategy.INCLUDE // required for flywayg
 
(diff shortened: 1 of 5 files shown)


braulio-dev:
Conceptually, I'm not a fan of passive skills behaving like triggered skills without an option to (soft) disable the (sub)skill. What I mean is it's probably gonna be annoying being a knight and the guy you just hit gets randomly pulled, that seems more like an active skill rather than a passive. It's probably more annoying than it is useful

Issue #2029, opened 2026-05-10
Item/Crafting registration is order based


StudentAlleg:
i.e.
Charcoal is registered after torch. Torch has a recipe that requires charcoal. This might break the recipe if using itemFactory.getFallbackItem(Material.CHARCOAL); to get the base item.

PR #2030, opened 2026-05-10
Fix being unable to craft torches

### LINKED ISSUE #2027: cannot craft torches

StudentAlleg:
Register torch as its own item to register a charcoal variant, as Minecraft:torch only has coal (probably some bug here on our end, as in vanilla both coal and charcoal work).

This also changes the order of registration, allowing coal to be used. Charcoal is still blocked because of #2029, which Rey said he was going to look at (or if he doesn't I can take a crack at it).

Fixes #2027


Mykindos:
@braulio-dev there must be a better way to keep all these alternate recipes intact

PR #2062, opened 2026-05-11
Created Cloth Recipe

8SIXSect:
## Describe your changes
- `Cloth` is used to craft a `Workbench` which is now an essential early game item; thus, there is a need for a recipe.
- `Cloth` can be crafted using 3 `Wool`


Mykindos:
The workbench isnt really essential when this works as one, just doesnt support blueprints (which arent even used yet)
image

Mykindos (approved):

8SIXSect:
> The workbench isnt really essential when this works as one, just doesnt support blueprints (which arent even used yet) image

Good point

PR #2120, opened 2026-05-15
Fix building jmh into /build/clans etc.

StudentAlleg:


StudentAlleg:
unneeded

PR #2128, opened 2026-05-17
Fix bug with treefeller

biglargegamer:
## Describe your changes
Right now if you change your item mid treefeller your axe won't lose the durability, this fixes that and makes the code better :).

### DIFF
--- a/core/src/main/java/me/mykindos/betterpvp/core/item/impl/interaction/TreeFellerInteraction.java
+++ b/core/src/main/java/me/mykindos/betterpvp/core/item/impl/interaction/TreeFellerInteraction.java
@@ -128,21 +128,11 @@ public class TreeFellerInteraction extends CooldownInteraction implements Displa
 
         UtilServer.callEvent(new TreeFellerEvent(player, initialLogLocation, initialLogType, blocksToFell));
 
+        final UUID playerUUID = player.getUniqueId();
         final ItemStack itemInMainHand = player.getInventory().getItemInMainHand();
-        for (int i = 0; i < blocksToFell.size(); i++) {
-            Block blockToFell = blocksToFell.get(i);
-            Location fellLocation = blockToFell.getLocation();
-            blocksBeingFelled.add(fellLocation);
-            UtilServer.runTaskLater(JavaPlugin.getPlugin(Core.class), () -> {
-                try {
-                    if (isBreakableLog(blockToFell, initialLogLocation)) {
-                        UtilBlock.breakBlock(player, blockToFell);
-                    }
-                } finally {
-                    blocksBeingFelled.remove(fellLocation);
-                }
-            }, (long) (i * 0.3));
-        }
+        final Iterator<Block> blockIterator = blocksToFell.iterator();
+        
+        treeFeller(playerUUID, blockIterator, initialLogLocation, itemInMainHand);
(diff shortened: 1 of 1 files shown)


core/src/main/java/me/mykindos/betterpvp/core/item/impl/interaction/TreeFellerInteraction.java
```diff
@@ -128,21 +128,11 @@ public double getCooldown(InteractionActor actor) {
 
         UtilServer.callEvent(new TreeFellerEvent(player, initialLogLocation, initialLogType, blocksToFell));
 
+        final UUID playerUUID = player.getUniqueId();
```
Mykindos, line 131: Why? Just pass Player into treeFeller directly, then you dont need to fetch the player again

core/src/main/java/me/mykindos/betterpvp/core/item/impl/interaction/TreeFellerInteraction.java
```diff
@@ -196,6 +186,32 @@ private void collectTree(@NotNull Block block, @NotNull Set<Location> visited, @
 
     }
 
+    private void treeFeller(UUID playerUUID, Iterator<Block> iterator, Location initialLogLocation, Material itemInMainHand) {
+        if (!iterator.hasNext()) return; 
+
+        Player player = Bukkit.getPlayer(playerUUID);
+        if (player == null || !player.isOnline()) return;
```
Mykindos, line 193: above

core/src/main/java/me/mykindos/betterpvp/core/item/impl/interaction/TreeFellerInteraction.java
```diff
@@ -196,6 +186,32 @@ private void collectTree(@NotNull Block block, @NotNull Set<Location> visited, @
...
+        Location fellLocation = nextBlock.getLocation();
+
+        blocksBeingFelled.add(fellLocation);
+        try {
+            if (isBreakableLog(nextBlock, initialLogLocation)) {
+                UtilBlock.breakBlock(player, nextBlock);
+            }
+        } finally {
+            blocksBeingFelled.remove(fellLocation);
+        }
```
Mykindos, line 208: blocksBeingFelled.add(fellLocation);

blocksBeingFelled.remove(fellLocation);

Does nothing with this list between them, why is it here?

core/src/main/java/me/mykindos/betterpvp/core/item/impl/interaction/TreeFellerInteraction.java
```diff
@@ -196,6 +186,32 @@ private void collectTree(@NotNull Block block, @NotNull Set<Location> visited, @
...
+
+        UtilServer.runTaskLater(JavaPlugin.getPlugin(Core.class), () -> {
+            treeFeller(playerUUID, iterator, initialLogLocation, itemInMainHand);
+        }, 1L);
```
Mykindos, line 212: Not really any different in the end, just forces a longer delay between each log break

Mykindos (changes requested):
Quite a few problems here, I feel like this PR could just be stripping down and simply adding the material check to support the existing system

PR #2137, opened 2026-05-17
Slight Assassin nerfs

StudentAlleg:
Assassin is the current most played class, almost more than the next 2 combined. Nerf Assassin reinforced armor to +8 overall (-1 from each piece). Very minor nerfs to Evade/Flash (high skill k/d and in top 3 used skills). Nerf backstab to 1.5 + 1.0L


Mykindos:
Need to analyze the data rather carefully here, we dont want to nerf something because its the most played but only if it is actually overpowered (this is where Kill death data versus OTHER kits is useful)

StudentAlleg:
image
Assassin is positive on every class, and over 1.75 for everyone

PR #2156, opened 2026-05-21
Add seeing what droptables an item is in to middle click

StudentAlleg:
Help players figure out where items come from by easily showing which droptable it exists in.


Mykindos:
A little concerned about the performance of this, currently items gui is a pool of 2 threads i believe and this does a LOT of iterations

StudentAlleg:
> A little concerned about the performance of this, currently items gui is a pool of 2 threads i believe and this does a LOT of iterations

It should be cached, the biggest problem is the first time, but I'd run into this problem with recipes on local (without this implementation)

PR #2164, opened 2026-05-23
Add getPlayer method and improve clan message formatting

TR4E:
- Client: add getPlayer method to retrieve Bukkit Player by UUID for easier player access
- ClanEventListener: unify clan and player name formatting with relation-based colors for clearer, consistent messaging across clan events; replace raw player and clan names with formatted components in messages and notifications; improve disband and alliance notifications to include relation-aware clan names; optimize online player messaging loops for clan events; enhance invite, join, leave, kick, promote, demote, alliance, trust, neutral, and enemy event messages with formatted names and relations; replace deprecated Bukkit.getPlayerExact with member/player getter methods


clans/src/main/java/me/mykindos/betterpvp/clans/clans/listeners/ClanEventListener.java
```diff
@@ -118,6 +95,24 @@ public ClanEventListener(final Clans clans, final ClanManager clanManager, final
...
+        return clanRelation.getPrimaryMiniColor() + clan.getName() + "<gray>";
+    }
+
+    private String getPlayerName(final ClanRelation clanRelation, final String playerName) {
+        return clanRelation.getPrimaryMiniColor() + playerName + "<gray>";
+    }
+
+    private String getPlayerName(final ClanRelation clanRelation, final Player player) {
+        return this.getPlayerName(clanRelation, player.getName());
+    }
```
Mykindos: Does not belong in a listener class

Mykindos (changes requested):

Mykindos (changes requested):
I found two behavior regressions in this patch that should be fixed before merge.

1. ClanEventListener#onClanSetCore removes the existing Material.BEDROCK guard before placing the new core. Previously we rejected placements when core.getSafest(highest) resolved to bedrock; now we always proceed, so the command can accept an invalid location and move/remove the core anyway.

2. ClanEventListener#onMemberPromote now formats the success message with clan.getName() instead of the promoted member's name. That means the actor sees You promoted <clan name>... instead of the player they actually promoted.

3. Dont use wildcard imports, you can change this in intelliJ settings

Once those are corrected, the rest of the message-formatting cleanup looks good to me.

clans/src/main/java/me/mykindos/betterpvp/clans/clans/listeners/ClanEventListener.java
```diff
@@ -942,7 +925,7 @@ public void onMemberPromote(final MemberPromoteEvent event) {
 
         this.clientManager.search().offline(member.getUuid()).thenAcceptAsync(result -> {
             result.ifPresent(client -> {
-                UtilMessage.simpleMessage(player, "Clans", "You promoted <aqua>%s<gray> to <yellow>%s<gray>.", client.getName(), member.getRank().getName());
+                UtilMessage.simpleMessage(player, "Clans", "You promoted %s to <yellow>%s</yellow>.", this.clanManager.getPlayerName(ClanRelation.SELF, clan.getName()), member.getRank().getName());
```
Mykindos: This message now formats `clan.getName()` as the promoted player name. That means the actor will see `You promoted <clan name>...` instead of the member they actually promoted. I think this should still use the promoted member/client name here.

clans/src/main/java/me/mykindos/betterpvp/clans/clans/listeners/ClanEventListener.java
```diff
@@ -905,11 +893,6 @@ public void onClanSetCore(final ClanSetCoreLocationEvent event) {
         Location highest = player.getLocation();
 
         final Block block = core.getSafest(highest).getBlock();
-        if(block.getType() == Material.BEDROCK) {
-            UtilMessage.simpleMessage(player, "Clans", "You cannot place the clan core here.");
-            return;
-        }
-
         core.removeBlock(); // Remove old core
```
Mykindos, line 924: This looks like the path where the old `Material.BEDROCK` guard used to apply. In the base commit we returned early when `core.getSafest(highest)` resolved to bedrock; after this change we go straight into `core.removeBlock()` and place the new core, so invalid bedrock placements are no longer rejected.

PR #2183, opened 2026-05-30
Tree Respawning

StudentAlleg:
Add a system that periodically respawns trees. This only happens in wilderness (no claimed chunk), only if there are not enough trees, and only if it is a valid spot to spawn a tree. Trees will spawn periodically over time, as they get queued into a limited 200 chunk queue. Only uses loaded chunks.

This is intended to replace trees that are felled by players in wilderness and not replanted. This allows players to then re-fell them to get wood and xp.

### DIFF
--- a/clans/src/main/java/me/mykindos/betterpvp/clans/world/TreeRespawnListener.java
+++ b/clans/src/main/java/me/mykindos/betterpvp/clans/world/TreeRespawnListener.java
@@ -1,0 +1,88 @@
+package me.mykindos.betterpvp.clans.world;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import me.mykindos.betterpvp.core.framework.updater.UpdateEvent;
+import me.mykindos.betterpvp.core.listener.BPvPListener;
+import me.mykindos.betterpvp.core.world.model.BPvPWorld;
+import org.bukkit.Bukkit;
+import org.bukkit.Chunk;
+import org.bukkit.World;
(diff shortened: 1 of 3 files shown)


clans/src/main/java/me/mykindos/betterpvp/clans/world/TreeRespawnListener.java
```diff
@@ -0,0 +1,88 @@
...
+        World world = Bukkit.getWorld(BPvPWorld.MAIN_WORLD_NAME);
+        if (world == null) return;
+
+        // Shuffle so every chunk gets a fair shot when the queue cap is hit
+        List<Chunk> chunks = Arrays.asList(world.getLoadedChunks());
+        Collections.shuffle(chunks);
+
+        for (Chunk chunk : chunks) {
+            treeRespawnManager.queueChunk(chunk);
+        }
```
Mykindos, line 72: Should probably store a weak reference to the chunk, as this will could / will issues when the chunk is unloaded                                                          and most likely memory issues as well

Mykindos (changes requested):

clans/src/main/java/me/mykindos/betterpvp/clans/world/TreeRespawnListener.java
StudentAlleg: that makes sense, done

Mykindos (approved):

PR #2187, opened 2026-05-31
Automate data collection

StudentAlleg:
Automate retrieving snapshots of aggregate player data over time. Add information from configs into db.

This is so we can create graphs/visualizations in grafana. Replaces current work I was doing in a sheets file for DPS/Damage analysis. I want to create visualizations of min/max DPS/Damage by weapon (with different runes, skills etc), TTK for various configurations, energy consumption, a builder to visualize this based off of a weapon/skill config, ability to look at all permutations. I want this information so that we can balance things like runes better (doing this exercise with some weapons showed how odd ancient sword is compared to legends).


Mykindos:
I have a hard time believing we cant already do this in grafana without storing even more data

StudentAlleg:
> I have a hard time believing we cant already do this in grafana without storing even more data
@Mykindos

Ok, short answer is yes, I still cannot do the client stats stuff (so skill stuff and class time played) without something being saved in the DB. But everything else could be done either with fetching the YAML files from somehere (i.e. github) via the Infinity plugin and the Role Matchup could be done with the data that is there.

Mykindos:
I think there are a few correctness issues worth tightening before this lands:

1. `GrafanaConfigSyncService.syncYamlConfig` only upserts the YAML leaves it currently sees, but it never removes rows for keys that were deleted or renamed. Since `grafana_config` is described as reflecting the live config, stale rows will survive forever after a reload/deploy and Grafana can keep joining against values that no longer exist. I'd make each file sync reconcile the full `(plugin, config_file)` set, e.g. delete keys not in the current leaf-key set in the same DB task, or replace that file's rows transactionally.

2. `takeSkillKdrSnapshot` builds the final snapshot from `skill_kills` as the base table. Any skill with deaths/time-played but zero kills is omitted entirely, which is probably exactly the kind of weak/underperforming skill you'd want visible in a balance dashboard. Consider building a base set from the union of kill/death/time skill names, or from the config-defined skills, then left-joining the aggregates.

StudentAlleg:
> I think there are a few correctness issues worth tightening before this lands:
> 
>     1. `GrafanaConfigSyncService.syncYamlConfig` only upserts the YAML leaves it currently sees, but it never removes rows for keys that were deleted or renamed. Since `grafana_config` is described as reflecting the live config, stale rows will survive forever after a reload/deploy and Grafana can keep joining against values that no longer exist. I'd make each file sync reconcile the full `(plugin, config_file)` set, e.g. delete keys not in the current leaf-key set in the same DB task, or replace that file's rows transactionally.
> 
>     2. `takeSkillKdrSnapshot` builds the final snapshot from `skill_kills` as the base table. Any skill with deaths/time-played but zero kills is omitted entirely, which is probably exactly the kind of weak/underperforming skill you'd want visible in a balance dashboard. Consider building a base set from the union of kill/death/time skill names, or from the config-defined skills, then left-joining the aggregates.

Done

PR #2215, opened 2026-07-19
fix(zone): prevent duplicate zone interact dispatch

TR4E:
- ZoneInteractionListener: add interactDeniedSet backed by WeakHashMap to skip queryAccess on the duplicate PlayerInteractEvent from the vanilla client's dual-packet UseItemOn/UseItem flow, while still setting useInteractedBlock to DENY on the second event


TR4E:
I am not 100% sure, if this should be implemented, but I've found an issue with like left-clicking or right-clicking container/gateway blocks where it'll send me a "You cannot use" message twice.

If this pull request ends up getting approved, that will be the culprit to check for this fix.

PR #2217, opened 2026-08-15
fix(combatlog): remove stale combat log entities

TR4E:
- CombatLog: mark spawned sheep with persistent combat log data and guard against null offline inventories before dropping items
- CombatLogManager: remove loaded living entities marked as combat logs when no active combat log matches them
- CombatLogListener: scan world and entity load events for stale combat log entities and remove them

### DIFF
--- a/core/src/main/java/me/mykindos/betterpvp/core/combat/combatlog/CombatLog.java
+++ b/core/src/main/java/me/mykindos/betterpvp/core/combat/combatlog/CombatLog.java
@@ -65,7 +65,7 @@ public class CombatLog {
         combatLogSheep.remove();
         CraftInventoryPlayer inventory = UtilInventory.getOfflineInventory(playerName, owner);
         if (inventory == null) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/me/mykindos/betterpvp/core/combat/combatlog/CombatLog.java
```diff
@@ -59,6 +64,10 @@ public void onClicked(Player player, WorldHandler worldHandler, OfflineMessagesH
 
         combatLogSheep.remove();
         CraftInventoryPlayer inventory = UtilInventory.getOfflineInventory(playerName, owner);
+        if (inventory == null) {
+            log.error("Failed to retrieve offline inventory for {} ({}) while processing combat log", playerName, owner);
```
Mykindos: needs .submit()

Mykindos (changes requested):

